PR #80, opened 2021-01-26
Clean up morphemic segmentations

loafofpiecrust:
- Put English glosses into single quotes, as is standard practice with
language materials.
- Put the morphemic segmentation in italics, differentiating it from the
layers above and below.
- Only show commentary icons at the "Analysis" level.
- Remove visual noise from morpheme buttons

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -14,6 +14,7 @@
   "private": true,
   "dependencies": {
     "@apollo/client": "^3.1.3",
+    "@fontsource/charis-sil": "^4.1.0",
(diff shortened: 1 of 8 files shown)


loafofpiecrust:
I think this may be a solid improvement on readability, especially for the 5-layer "Segmentation" and "Analysis" modes. Y'all can see a [preview here](https://deploy-preview-80--dailp.netlify.app/documents/dd1). I'd love to hear any thoughts @jgbourns @LN7-cyber @GracefulLemming

PR #108, opened 2021-05-06
Added anchors for source layer in all view types

### LINKED ISSUE #74: Anchor links for each word in a document
I would like to add an anchor link to each word to make it easier for users to send a link to a specific form to someone else.

-  in modes with word-by-word breakdown: anchor link per word
-  otherwise an anchor link for each translated block

GracefulLemming:
Adds anchor tags to source layer in annotated forms and story mode sections. 
closes #74

### DIFF
--- a/website/src/segment.tsx
+++ b/website/src/segment.tsx
@@ -109,35 +109,40 @@ export const AnnotatedForm = (
     const showSegments = p.viewMode >= ViewMode.Segmentation
     const translation = p.segment.englishGloss.join(", ")
     return (
-      <div className={wordGroup} id={`w${p.segment.index}`}>
-        <div className={syllabaryLayer} lang="chr">
-          {p.segment.source}
-          {p.segment.commentary && p.viewMode >= ViewMode.Pronunciation && (
-            <WordCommentaryInfo commentary={p.segment.commentary} />
+      <a className={anchor} id={`w${p.segment.index}`}>
+        <div className={wordGroup} id={`w${p.segment.index}`}>
+          <div className={syllabaryLayer} lang="chr">
+            {p.segment.source}
+            {p.segment.commentary && p.viewMode >= ViewMode.Pronunciation && (
+              <WordCommentaryInfo commentary={p.segment.commentary} />
+            )}
+          </div>
+          {p.segment.simplePhonetics ? (
+            <div>{p.segment.simplePhonetics}</div>
+          ) : (
+            <br />
           )}
+          {showSegments ? (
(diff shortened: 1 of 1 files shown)


website/src/segment.tsx
```diff
@@ -109,35 +109,40 @@ export const AnnotatedForm = (
     const showSegments = p.viewMode >= ViewMode.Segmentation
     const translation = p.segment.englishGloss.join(", ")
     return (
-      <div className={wordGroup} id={`w${p.segment.index}`}>
-        <div className={syllabaryLayer} lang="chr">
-          {p.segment.source}
-          {p.segment.commentary && p.viewMode >= ViewMode.Pronunciation && (
-            <WordCommentaryInfo commentary={p.segment.commentary} />
+      <a className={anchor} id={`w${p.segment.index}`}>
+        <div className={wordGroup} id={`w${p.segment.index}`}>
```
loafofpiecrust, line 113: I would consider merging these into one <a> tag or probably instead of making the whole word a link, add a small anchor element next to the word that only shows when it's hovered or focused. I've seen this on many sites where a `#` symbol to the left of a header represents the anchor link.

website/src/segment.tsx
```diff
@@ -109,35 +109,40 @@ export const AnnotatedForm = (
...
-        ) : null}
-        {translation.length ? <div>&lsquo;{translation}&rsquo;</div> : <br />}
-      </div>
+      </a>
     )
   } else {
     return (
-      <span className={plainSyllabary} id={`w${p.segment.index}`} lang="chr">
+      <a className={anchor} id={`w${p.segment.index}`}>
+        <span className={plainSyllabary} id={`w${p.segment.index}`} lang="chr">
```
loafofpiecrust, line 141: Does making each word within a paragraph a link seem overwhelming to you?

website/src/segment.tsx
```diff
@@ -411,3 +416,6 @@ const annotationSection = css`
 const storySection = css`
   flex-flow: row wrap;
 `
+const anchor =css`
+  color:black;
+`
```
loafofpiecrust, line 421: Check the theme.ts file for project wide color and font definitions to use instead of literal values. I'd say if you want a link that doesn't look like one, then it's not the right approach.

loafofpiecrust:
Checking in: have you made any progress on this one? Even if it's small, it doesn't have to be completely finished for us to take a second pass at it.

GracefulLemming:
This PR has been stale for quite some time. If users are still interested, we can revisit in the coming months.

PR #143, opened 2022-02-01
[DP-77] Replacing empty breaks with solid breaks in document parsing

KushP10:
Added black page breaks in places of empty spaces within document viewer

### DIFF
--- a/website/src/segment.css.ts
+++ b/website/src/segment.css.ts
@@ -146,6 +146,7 @@ const annotationSectionBase = style([
     width: "100%",
     position: "relative",
     marginBottom: vspace.half,
+    display: "flex",
   },
 ])
 
@@ -185,6 +186,6 @@ export const lineBox = style([
     textAlign: "start",
     padding: ".75em",
     marginLeft: 0,
-    paddingLeft: "0em"
+    paddingLeft: "0em",
   },
-]) 
+])
--- a/website/src/segment.css.ts
+++ b/website/src/segment.css.ts
@@ -167,3 +167,15 @@ export const infoIcon = style({
 })
 
 export const linkSvg = style({ fill: colors.link })
+
+export const fillerLine = style([
+  {
+    "@media": {
+      [mediaQueries.medium]: {
+        borderBottom: "1px solid black",
+        padding: 0,
+        ...margin(vspace.one, "3rem", vspace.one, 0),
+      },
+    },
+  },
+])
--- a/website/src/segment.css.ts
+++ b/website/src/segment.css.ts
@@ -146,7 +146,12 @@ const annotationSectionBase = style([
     width: "100%",
     position: "relative",
     marginBottom: vspace.half,
-    display: "flex",
+    "@media": {
+      [mediaQueries.medium]: {
+        display: "flex",
+        flexFlow: "row wrap",
+      },
+    },
   },
 ])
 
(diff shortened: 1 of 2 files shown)


website/src/segment.tsx
```diff
@@ -190,16 +190,15 @@ const MorphemicSegmentation = (p: {
...
 }) => {
-  // If there is no segmentation, return two line breaks for the
+  // If there is no segmentation, return a hard break for the
   // morphemic segmentation and morpheme gloss layers.
   if (!p.segments?.length) {
     return (
       <>
-        <br />
-        <br />
+        <hr className={css.fillerLine}/>
```
loafofpiecrust, line 198: I think we should make a component for this `<hr/>` tag. That might look like:

```ts
const FillerLine = () => (<hr className={css.fillerLine}/>)
```

And then we also want to retain the vertical rhythm of our word annotations, so make sure to include two here, not just one. i.e.

```ts
<FillerLine />
<FillerLine />
```

website/src/segment.css.ts
```diff
@@ -167,3 +167,15 @@ export const infoIcon = style({
...
+  {
+    "@media": {
+      [mediaQueries.medium]: {
+        borderBottom: "1px solid black",
+        padding: 0,
+        ...margin(vspace.one, "3rem", vspace.one, 0),
+      },
+    },
+  },
+])
```
loafofpiecrust, line 181: We might want to run `prettier` on a commit hook or something but for now, would you mind turning on "format on save" in your editor? If you're using VSCode it has built-in support for `prettier`. That'll automatically fix this line ending thing and some extra spaces you have elsewhere.

website/src/segment.css.ts
```diff
@@ -167,3 +167,15 @@ export const infoIcon = style({
...
 
 export const linkSvg = style({ fill: colors.link })
+
+export const fillerLine = style([
+  {
+    "@media": {
+      [mediaQueries.medium]: {
+        borderBottom: "1px solid black",
+        padding: 0,
+        ...margin(vspace.one, "3rem", vspace.one, 0),
```
loafofpiecrust, line 177: This will only work on screen sizes larger than mobile. Make sure to try out the mobile version either using the "responsive" feature in the browser dev tools or just shrinking your browser window manually. Then for mobile styles, just add those to the top level. For example:

```ts
style({
  // Mobile and all size styles go here...
  borderBottom: "1px solid black",
  "@media": {
    [mediaQueries.medium]: {
      // Tablet and desktop styles go here...
    }
  }
})
```

loafofpiecrust: Also, a small note on styling, we don't want to use a hardcoded color like `black` here. We probably want the line to be the same color as our text, whatever color that may be. So we can use `borderColor: colors.text` instead.

website/src/segment.tsx
```diff
@@ -180,6 +184,13 @@ const WithTooltip = (p: {
   </Tooltip>
 )
 
+const FillerLine = () => (
+  <div className={css.lineBox}>
+    {" "}
+    <hr className={css.fillerLine} />{" "}
```
loafofpiecrust, line 190: Very last thing, we can get rid of these `{" "}` which are just explicit spaces that got added when reformatting, because we don't need any explicit extra space characters here.

website/src/segment.css.ts
```diff
@@ -146,6 +146,7 @@ const annotationSectionBase = style([
     width: "100%",
     position: "relative",
     marginBottom: vspace.half,
+    display: "flex",
```
loafofpiecrust, line 149: It looks like this change actually totally messed up the mobile view. Make sure to test that stuff first even if I suggest it 🤣 . A better solution might be something like this:
```ts
"@media": {
  [mediaQueries.medium]: {
    display: "flex",
    flexFlow: "row wrap",
  }
}
```

loafofpiecrust (approved):

PR #147, opened 2022-02-14
[DP-61] Adding a word panel

patelShas:
This adds a word panel to the side of the translation tab that can be opened up by clicking on a word.

image

This Word panel remains on the side even as the main text is scrolled past or before, and a rectangle points out what word has been selected. The word panel has an "X" that will close the panel, but clicking on another word will change what word's details the panel is showing. 

image

On a smaller screen or mobile device, the word panel will appear on a drawer from the right side. This drawer can be closed by either clicking the "X", or simply clicking off the drawer.

### DIFF
--- a/website/src/pages/documents/document.page.tsx
+++ b/website/src/pages/documents/document.page.tsx
@@ -4,6 +4,12 @@ import React, { Dispatch, SetStateAction, useState } from "react"
 import { isMobile } from "react-device-detect"
 import { Helmet } from "react-helmet"
 import Sticky from "react-stickynode"
+import {
+  Dialog,
+  DialogBackdrop,
+  DialogDisclosure,
+  useDialogState,
+} from "reakit/Dialog"
 import { Tab, TabList, TabPanel } from "reakit/Tab"
 import { DocumentAudio } from "src/audio-player"
 import { Breadcrumbs } from "src/breadcrumbs"
@@ -11,6 +17,7 @@ import { Button } from "src/components"
 import Link from "src/components/link"
 import * as Dailp from "src/graphql/dailp"
 import Layout from "src/layout"
+import { drawerBg, navButton, navDrawer } from "src/menu.css"
 import { ExperiencePicker, selectedMode, selectedPhonetics } from "src/mode"
 import { MorphemeDetails } from "src/morpheme"
 import {
@@ -20,7 +27,6 @@ import {
 } from "src/routes"
 import { useScrollableTabState } from "src/scrollable-tabs"
 import { AnnotatedForm, Segment } from "src/segment"
-import { WordPanel, wordPanelDetails } from "src/word-panel"
 import {
   BasicMorphemeSegment,
   PhoneticRepresentation,
@@ -28,10 +34,9 @@ import {
   ViewMode,
   tagSetForMode,
 } from "src/types"
+import { WordPanel, WordPanelDetails } from "src/word-panel"
 import PageImages from "../../page-image"
 import * as css from "./document.css"
-import { Dialog, DialogBackdrop, DialogDisclosure, useDialogState } from "reakit/Dialog"
-import { navButton, drawerBg, navDrawer } from "src/menu.css"
 
 enum Tabs {
   ANNOTATION = "annotation-tab",
@@ -135,10 +140,11 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
     setDialogOpen(true)
   }
 
-  const dialog = useDialogState({ animated: true})
-  const [wordPanelContents, setwpContents] = useState<Dailp.FormFieldsFragment | null>(null)
-  const setWordPanelContents = (content: Dailp.FormFieldsFragment | null) => {
-    setwpContents(content)
+  const dialog = useDialogState({ animated: true })
+  const [selectedWord, setSelectedWord] =
+    useState<Dailp.FormFieldsFragment | null>(null)
+  const selectAndShowWord = (content: Dailp.FormFieldsFragment | null) => {
+    setSelectedWord(content)
     if (content) {
       dialog.show()
     } else {
@@ -146,8 +152,8 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
     }
   }
   let wordPanelInfo = {
-    currContents: wordPanelContents,
-    setCurrContents: setWordPanelContents,
+    currContents: selectedWord,
+    setCurrContents: selectAndShowWord,
   }
 
   const [phoneticRepresentation, _setPhoneticRepresentation] =
@@ -181,7 +187,6 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
         </DialogContent>
       </DialogOverlay>
 
-      
       <DialogBackdrop {...dialog} className={drawerBg}>
         <Dialog
           {...dialog}
@@ -206,9 +211,9 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
       </p>
 
       <SolidSticky top="#document-tabs-header">
-        <ExperiencePicker onSelect={setExperienceLevel}/>
+        <ExperiencePicker onSelect={setExperienceLevel} />
         {/*<PhoneticsPicker onSelect={setPhoneticRepresentation} />*/}
-      </SolidSticky> 
+      </SolidSticky>
 
       <section className={css.contentContainer}>
         <article className={css.annotationContents}>
@@ -219,11 +224,11 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
               openDetails,
               tagSet,
               phoneticRepresentation,
-              wordPanelInfo,
+              wordPanelDetails: wordPanelInfo,
             }}
           />
         </article>
-        {wordPanelContents ? (
+        {selectedWord ? (
           <div className={css.contentSection2}>
             <WordPanel
               segment={wordPanelInfo.currContents}
@@ -233,10 +238,7 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
               tagSet={tagSet}
             />
           </div>
-        ) : (
-          null
-        )}
-
+        ) : null}
       </section>
     </>
   )
@@ -248,14 +250,14 @@ const DocumentContents = ({
   openDetails,
   tagSet,
   phoneticRepresentation,
-  wordPanelInfo,
+  wordPanelDetails,
 }: {
   doc: Document
   experienceLevel: ViewMode
   tagSet: TagSet
   openDetails: (morpheme: any) => void
   phoneticRepresentation: PhoneticRepresentation
-  wordPanelInfo: wordPanelDetails
+  wordPanelDetails: WordPanelDetails
 }) => {
   let morphemeSystem = Dailp.CherokeeOrthography.Learner
   if (experienceLevel === ViewMode.AnalysisDt) {
@@ -283,7 +285,7 @@ const DocumentContents = ({
           translations={seg.translation as Dailp.TranslationBlock}
           pageImages={doc.pageImages?.urls!}
           phoneticRepresentation={phoneticRepresentation}
-          wordPanelInfo={wordPanelInfo}
+          wordPanelDetails={wordPanelDetails}
         />
       ))}
       {docContents.forms?.map((form, i) => (
@@ -296,7 +298,7 @@ const DocumentContents = ({
           phoneticRepresentation={phoneticRepresentation}
           translations={null}
           pageImages={doc.pageImages?.urls!}
-          wordPanelInfo={wordPanelInfo}
+          wordPanelDetails={wordPanelDetails}
         />
       ))}
     </>
--- a/website/src/pages/documents/document.page.tsx
+++ b/website/src/pages/documents/document.page.tsx
@@ -1,6 +1,6 @@
 import { DialogContent, DialogOverlay } from "@reach/dialog"
 import "@reach/dialog/styles.css"
-import React, { Dispatch, SetStateAction, useState } from "react"
+import React, { Dispatch, SetStateAction, useEffect, useState } from "react"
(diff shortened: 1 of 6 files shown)


website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,32 @@
+import { style } from "@vanilla-extract/css"
+import sprinkles from "./sprinkles.css"
+
+export const wholePanel = style({
+    position: "sticky",
```
loafofpiecrust, line 5: I never knew about this value for `position`! Looks like it has [pretty good browser support as long as you don't use it on table elements](https://caniuse.com/?search=sticky), but only if we don't do IE support. Looks like there's [a polyfill](https://github.com/wilddeer/stickyfill) if we do need to support IE. Great find!

website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,32 @@
...
+    lineHeight: 1.5,
+    boxSizing: "border-box",
+    border: "transparent",
+    cursor: "pointer",
+    fontSize: "16px",
+    borderRadius: "0.25rem",
+    ":before":{
+        display: "inline-block",
+        margin: "4px",
+    },
```
loafofpiecrust, line 19: I'm not certain this works? With [vanilla-extract](https://vanilla-extract.style/documentation/styling-api/) you normally need to use
```ts
selectors: {
  "&:before": {
    ...
  }
}
```

website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,32 @@
...
+    position: "absolute",
+    top: "5px",
+    right: "5px",
```
loafofpiecrust, line 22: Use `vspace` and `hspace` for padding, margins, and absolute positioning.

website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,32 @@
...
+    position: "sticky",
+    top: 150,
+})
+
+export const wordPanelButton = style({
+    lineHeight: 1.5,
+    boxSizing: "border-box",
+    border: "transparent",
+    cursor: "pointer",
+    fontSize: "16px",
```
loafofpiecrust, line 14: Use a relative [font size in `rem` units](https://www.sitepoint.com/understanding-and-using-rem-units-in-css/), for example this might just be `1.1rem` if you want bigger than normal body text or `0.9rem` if you want a bit smaller.

website/src/word-panel.tsx
```diff
@@ -0,0 +1,87 @@
...
+                            index={p.segment.audioTrack.index}
+                            parentTrack=""
+                            resourceUrl={p.segment.audioTrack.resourceUrl}
+                            startTime={p.segment.audioTrack.startTime}
+                            />
+                        )}
+                        </div>
+                    </>
+                    ) : (
+                    (p.segment.audioTrack && (
```
loafofpiecrust, line 52: You can use `x ? a : b` instead of `(x && a) || b` here.

website/src/word-panel.tsx
```diff
@@ -0,0 +1,87 @@
...
+        const translation = p.segment.englishGloss.join(", ")
+        return (
+            <div className={css.wholePanel}>
+                
+                <div 
+                className={css.wordPanelContent}>
+                    <Button className={css.wordPanelButton} onClick={() => p.setContent(null)}>X</Button>
+
+                    <h2>{p.segment.source}</h2> 
+                    {p.segment.simplePhonetics ? (
```
loafofpiecrust, line 35: Since there's like three different options here based on two conditions, I would extract this out into a flatter structure. Maybe like

```ts
let wordBreakdown = null
if (x) {
  wordBreakdown = ...
} else if (y) {
  wordBreakdown = ...
}

return (
  <div>
    {wordBreakdown}
  </div>
)
```

loafofpiecrust (commented):
Thanks for the full description and screenshots, super helpful!

- I'd like to see the details panel fill closer to the viewport height. 
- Let's also make the box around the currently selected word go around all of it's contents. Right now it's just around the first line, but I think surrounding the whole thing makes sense.
- Let's use an icon for the close button. I think we are for the `MorphemeDialog` component, so we should do something similar. Icon button with an `aria-label` that says "Close word details" or similar.
- It looks like, at least on mobile, that the fonts for the interlinear gloss aren't quite right. Make sure to include that styling, it really matters for the word analysis b/c that often has certain accents.
- Now that I'm looking at it, we should make sure we provide some guideposts especially while this panel lacks much info. So, would you mind changing the header to "Selected Word" and putting the Cherokee syllabary of the word below that?
- On mobile, the border shouldn't sit inside the sidebar. Either remove the border on mobile or remove the margin separating it from the edge.

website/src/pages/documents/document.page.tsx
```diff
@@ -132,6 +135,21 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
     setDialogOpen(true)
   }
 
+  const dialog = useDialogState({ animated: true})
+  const [wordPanelContents, setwpContents] = useState<Dailp.FormFieldsFragment | null>(null)
```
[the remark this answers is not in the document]
patelShas: Yeah, those names would be better fitting for the functions. I'll make the changes.

website/src/pages/documents/document.page.tsx
```diff
@@ -193,12 +248,14 @@ const DocumentContents = ({
...
   tagSet,
   phoneticRepresentation,
+  wordPanelInfo,
 }: {
   doc: Document
   experienceLevel: ViewMode
   tagSet: TagSet
   openDetails: (morpheme: any) => void
   phoneticRepresentation: PhoneticRepresentation
+  wordPanelInfo: wordPanelDetails
```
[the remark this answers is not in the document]
patelShas: I wasn't sure how else I could pass on the information on the Word Panel, since the word panel itself is on the translation tab, and the segments I want to get it passed to are in the DocumentContents in the translation tab. It seemed easier to pass it down in an object than to pass down each thing individually.

website/src/segment.tsx
```diff
@@ -85,13 +89,22 @@ export const AnnotatedForm = (
...
 
     return (
       <div className={css.wordGroup} id={`w${p.segment.index}`}>
-        <div className={css.syllabaryLayer} lang="chr">
+        
+        
+        <div 
+        className={wordCSS} 
+        lang="chr" 
+        onClick={() => p.wordPanelInfo.setCurrContents(p.segment)}>
```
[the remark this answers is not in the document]
patelShas: Yeah, I wasn't aware of format on save, so I'll enable and use on all changes.

website/src/word-panel.css.tsx
patelShas: Yeah, I wasn't quite sure about how to write this section. I must have left that part out.

loafofpiecrust: Just checked the docs again, and for simple pseudo-selectors you don't need to nest them under "selectors", so the code you wrote should work just fine.

website/src/pages/documents/document.page.tsx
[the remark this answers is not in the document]
loafofpiecrust: Yeah my feeling is if we are (almost) always passing these fields together and there are more than 2, then a sub-object makes sense. We can keep it for now at least.

patelShas:
Update and made several changes to the word panel.

image

Notable changes:
- The word panel has been brought closer to the top
- The box for the selected word now circles the entire form, not just the word
- The close button now has an icon rather than just "X"
- "Selected Word:" added to the word panel

image

And on mobile:
- The word panel no longer has a box outline inside the drawer
- The font is the same as on desktop and the main page

patelShas:
image

The selected word is now more distinctive.

website/src/pages/documents/document.page.tsx
```diff
@@ -172,17 +215,31 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
...
-          {...{
-            experienceLevel,
-            doc,
-            openDetails,
-            tagSet,
-            phoneticRepresentation,
-          }}
-        />
-      </article>
+      <section className={css.contentContainer}>
```
loafofpiecrust, line 218: I wouldn't use the `section` semantic element here because you are just grouping two things together, it's not really a thematic section of content. I would probably use a `div` here.

website/src/word-panel.tsx
```diff
@@ -0,0 +1,95 @@
...
+            resourceUrl={p.segment.audioTrack.resourceUrl}
+            startTime={p.segment.audioTrack.startTime}
+          />
+        </div>
+      )
+    }
+
+    return (
+      <div className={css.wholePanel}>
+        <div className={css.wordPanelContent}>
```
loafofpiecrust, line 69: So looks like you have the word panel nested in three `div`s. Two here, and one in `document.page.tsx`. Just for layout you almost never need three directly nested elements, so let's collapse the one in `document.page.tsx`.

website/src/segment.css.ts
```diff
@@ -63,7 +62,15 @@ export const wordGroup = style([
     "@media": {
       [mediaQueries.medium]: {
         display: "inline-block",
-        padding: 0,
+      }
+    }
+  },
+])
+
+export const wordGroupSelection = styleVariants({
```
loafofpiecrust, line 70: Not sure where in the styling this happens, but words get all messed up when you select them. [I'm looking here](https://deploy-preview-147--dailp.netlify.app/documents/dd6); when I select one word, then select another they shift around in the document. We really can't have that, we need the words to sit still when we (de)select them.

Looks like it might be this: You can't change the padding or margin between the selected and unselected states because that will change the layout flow.

website/src/segment.tsx
```diff
@@ -85,13 +89,21 @@ export const AnnotatedForm = (
...
     const translation = p.segment.englishGloss.join(", ")
 
     return (
-      <div className={css.wordGroup} id={`w${p.segment.index}`}>
-        <div className={css.syllabaryLayer} lang="chr">
+      <div className={wordCSS} id={`w${p.segment.index}`}>
+        <div
+          className={css.syllabaryLayer}
+          lang="chr"
+          onClick={() => p.wordPanelDetails.setCurrContents(p.segment)}
```
loafofpiecrust, line 105: As I play with it, I'd like users to be able to highlight the text of a word without changing their selection state. So, can we only select the word if you click on the information icon in the top-right corner of each word? That's how the designs are envisioned.

For this, you should also change the cursor style for the info icon to be `select`. And please add what currently shows as a tooltip when you hover on the info icon to the word details panel, below the multi-line translation. And then get rid of the tooltip.

website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,39 @@
+import { style } from "@vanilla-extract/css"
+import sprinkles, { hspace, mediaQueries, theme, vspace } from "./sprinkles.css"
+
+export const wholePanel = style({
```
loafofpiecrust, line 4: I think I mentioned this before, but we do want the panel to fill to the bottom of the screen. This way, it doesn't change height if a word has more information than another.

website/src/word-panel.css.tsx
```diff
@@ -0,0 +1,39 @@
...
+  position: "sticky",
+  top: 125,
+})
+
+export const wordPanelButton = style({
```
loafofpiecrust, line 9: You can look at the close button for the `MorphemeDialog` which already has some spacing styles & clears out the background so it's just the icon. I think that styling will look better here.

website/src/word-panel.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <Button
+            className={css.wordPanelButton}
+            onClick={() => p.setContent(null)}
+            aria-label="Close Word Details"
+          >
+            <MdClose size={32} />
+          </Button>
+          <h2>Selected word:</h2>
+          <h1>{p.segment.source}</h1>
```
loafofpiecrust, line 78: Swap the header types here, use an `h1` for "Selected Word" and an `h2` for the word source text, since that follows "main header" and "subheader".

website/src/segment.tsx
patelShas: That makes sense, but I'd like to clarify what you mean about the cursor style for the info icon. As far as I can tell, select is not a single cursor style but a category where there are multiple. The most similar I think is `text` which shows the cursor as an I-beam. Even still, I don't quite get why we would use this, as it usually shows text selection, and the info icon is simply something we click on. 

Do you mean having this text selection icon for the whole word except the info icon, or some other cursor style?

patelShas:
image

Added the latest changes:

- Words now all have info icons. The word panel is now opened by clicking this info icon, not the word's details.
- Word commentary is now displayed on the Word panel after other details.
- Padding is adjusted so that selecting different words doesn't change how they are laid out on the page.
- The word panel is extended to fill out more of the page.
- The heading and close buttons have both been changed in appearance.

patelShas:
Just changed height of the word panel to scale a bit more with the screen's actual size.

loafofpiecrust:
It looks like you're assuming words are uniquely identified by their syllabary layer, which is not true. Make sure to check the word index instead, so that you don't get duplicate selections like this:

Screen Shot 2022-02-28 at 11 10 10 AM

patelShas:
- Fixed up the fonts on the word panel to be more correct.
- Fixed an issue where if a word appeared multiple times in the document, all occasions would be highlighted when the word was selected. Now, only the one clicked will be highlighted.

patelShas:
Fixed another issue where on smaller screen or mobile devices (where the word panel appeared as a drawer from the left) the selected word would still be highlighted after the word panel is closed. Closing the word panel now removes the highlight.

patelShas:
- Made it so the word panel updates immediately when experience level/analysis is changed
- Allowed detailed morpheme dialog to appear without closing the word panel
- The word panel will close when switching to story mode

loafofpiecrust (approved):

PR #150, opened 2022-03-08
[DP-128] Adding Collapsible Details Panels to the Word Panel

patelShas:
image
The Word Panel now has its details show in collapsible panels.
image
The right arrow also changes based on whether the details are open are not. 
image
If there are more details than can fit on the word panel, the word panel gets a scroll bar.

Some other notes:
- The last collapsible panel for commentary is only visible for when the word has commentary. Otherwise, only the first two appear.
- I've also added a new element function that will create collapsible panels based on Title, Icon, and Content. This is so that any future details can be added to the word panel.

### DIFF
--- a/website/src/word-panel.css.tsx
+++ b/website/src/word-panel.css.tsx
@@ -4,9 +4,13 @@ import sprinkles, {
   colors,
   hspace,
   mediaQueries,
+  radii,
   theme,
+  thickness,
   vspace,
 } from "./sprinkles.css"
+
+const wordPanelPadding = "8px"
 
 export const cherHeader = style({
   color: colors.headings,
@@ -15,21 +19,61 @@ export const cherHeader = style({
 
 export const wordPanelButton = styleVariants({
   basic: [closeButton],
+  colpright: [
+    closeButton,
+    {
+      position: "relative",
+      float: "right",
+      top: "0px",
+    },
+  ],
+  colpleft: [
+    closeButton,
+    {
+      position: "relative",
+      float: "left",
+      top: "0px",
+      marginRight: wordPanelPadding,
+      left: "0px",
+    },
+  ],
+})
+
+export const collPanelContent = style({
+  padding: wordPanelPadding,
+  marginBottom: "0px",
+})
+
+export const collPanelButton = style({
+  flex: 1,
+  backgroundColor: "rgba(0, 0, 0, 0)",
+  border: "none",
+  textAlign: "left",
+  padding: wordPanelPadding,
+})
+
+export const collPanel = style({
+  width: "inherit",
+  display: "flex",
+  flexDirection: "column",
+  border: "1px solid #ddd",
 })
 
 export const wordPanelContent = style({
   position: "sticky",
-  top: 125,
-  height: "calc(100vh - 150px)",
+  top: 100,
+
   border: "none",
   borderRadius: "4px",
-  padding: "8px",
+  padding: wordPanelPadding,
   fontFamily: theme.fonts.body,
   "@media": {
     [mediaQueries.medium]: {
       border: "1px solid #ddd",
+      height: "calc(100vh - 125px)",
     },
   },
+  overflowY: "auto",
 })
 
 export const audioContainer = style({ paddingLeft: "40%" })
--- a/website/src/word-panel.tsx
+++ b/website/src/word-panel.tsx
@@ -1,5 +1,7 @@
 import React from "react"
-import { MdClose } from "react-icons/md"
+import { AiFillCaretDown, AiFillCaretUp } from "react-icons/ai"
+import { IoEllipsisHorizontalCircle } from "react-icons/io5"
+import { MdClose, MdNotes, MdRecordVoiceOver } from "react-icons/md"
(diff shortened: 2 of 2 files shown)


website/src/word-panel.css.tsx
```diff
@@ -4,32 +4,76 @@ import sprinkles, {
...
+    },
+  ],
+})
+
+export const collPanelContent = style({
+  padding: wordPanelPadding,
+  marginBottom: "0px",
+})
+
+export const collPanelButton = style({
```
loafofpiecrust, line 47: Let's apply the header font to this style, since these are kinda headings of small sections.

website/src/word-panel.css.tsx
```diff
@@ -4,32 +4,76 @@ import sprinkles, {
...
 export const wordPanelContent = style({
   position: "sticky",
-  top: 125,
-  height: "calc(100vh - 150px)",
+  top: 100,
+
   border: "none",
   borderRadius: "4px",
-  padding: "8px",
+  padding: wordPanelPadding,
```
loafofpiecrust, line 68: Can we actually remove this padding so that the collapsible sections fill the width of the details panel? And then we can put the `h1` and `h2` inside of a `<header>` element which has this padding instead, to make sure those aren't up against the panel edge.

loafofpiecrust (commented):
Looking really good! Thanks for anticipating some things we want, like adding icons to the sections and splitting out "Phonetics" into its own grouping. I may have more comments later but here's most of my thoughts. Thanks!

website/src/word-panel.tsx
```diff
@@ -75,19 +77,77 @@ export const WordPanel = (p: {
...
+
+const CollapsiblePanel = (p: {
+  title: String
+  content: JSX.Element
+  icon: JSX.Element // Note : this is supposed to be an IconType
+}) => {
+  const disclosure = useDisclosureState({ visible: true })
+  return (
+    <div className={css.collPanel}>
+      <Disclosure {...disclosure} className={css.collPanelButton}>
```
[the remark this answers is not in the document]
patelShas: I didn't think of that but yeah. The header for the panel and the aria label would be the same, but I don't think that will be an issue.

website/src/word-panel.css.tsx
patelShas: Do you think we should also change the font's color as well, since they are headers? Or just make sure its the right font but not change the color as to not muddle the separation between content and the word Panel's header.

website/src/word-panel.css.tsx
patelShas: Yeah! I didn't know about the header element before hand. I'll implement it.

website/src/word-panel.tsx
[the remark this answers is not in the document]
loafofpiecrust: Yeah so `aria-label` is only for things like screen readers and it's the same as the section name but without the two icons potentially adding noise, or making the browser recompute the label because the container contents change (when the arrow icon switches).

website/src/word-panel.css.tsx
loafofpiecrust: I think we can leave the color as-is since it's not in exactly the same layout as a header.

patelShas:
image

Collapsible panel has only borders on top and between each other.

image

Mobile/small screen version has been made to look similar too.

loafofpiecrust:
Looking really slick! The only thing left is to add `aria-label` to the group headings and an `aria-label` to the close button. And then we can merge this sucker!

patelShas:
> Looking really slick! The only thing left is to add `aria-label` to the group headings and an `aria-label` to the close button. And then we can merge this sucker!

The heading of the collapsible panel and the open/close button are the same element. Would we give them the same aria-label then or give a special separate label to the changing icon?

loafofpiecrust:
Sorry, by the "close button" I meant the big X at the top-right of the word details panel. We should give each section heading one label for the `Disclosure` element, no separate labels for the icons there.

website/src/word-panel.tsx
```diff
@@ -137,7 +137,11 @@ const CollapsiblePanel = (p: {
   const disclosure = useDisclosureState({ visible: true })
   return (
     <div className={css.collPanel} aria-label={p.title}>
-      <Disclosure {...disclosure} className={css.collPanelButton}>
+      <Disclosure
+        {...disclosure}
+        className={css.collPanelButton}
+        aria-label={`${p.title} header`}
```
loafofpiecrust, line 143: Let's just use `p.title` here, screen readers will already get an indication that this is a button which opens more details.

patelShas: That feels redundant though since the whole element, the full collapsible panel, already has an aria-label that is p.title. Would we have to remove that label for the whole panel or change it to something else?

loafofpiecrust: Oh I'm sorry, I didn't see that! I think that common practice is to label the `Disclosure` button and not label the div we are putting around it.

patelShas: Ah! okay.

loafofpiecrust (approved):

PR #153, opened 2022-03-24
added styling and separators in addition to descriptors

KushP10:
image
image

### DIFF
--- a/website/src/word-panel.css.tsx
+++ b/website/src/word-panel.css.tsx
@@ -92,5 +92,6 @@ export const tableCells = style({
   border: "4px solid transparent",
   fontFamily: theme.fonts.body,
   padding: "0px",
+  paddingRight: "3px",
   wordWrap: "break-word",
 })
--- a/website/src/word-panel.css.tsx
+++ b/website/src/word-panel.css.tsx
@@ -82,10 +82,14 @@ export const wordPanelHeader = style({
 
 export const audioContainer = style({ paddingLeft: "40%" })
 
-export const table = style({
+export const tableContainer = style({
+  border: "0px solid transparent",
   width: "max-content",
-  borderBottom: "0px solid #ddd",
+})
+
+export const tableCells = style({
+  border: "0px solid transparent",
+  borderRight: "5px solid transparent",
   fontFamily: theme.fonts.body,
-  padding: "0px",
-  borderRadius: "0px",
+  padding: "3px",
 })
--- a/website/src/word-panel.css.tsx
+++ b/website/src/word-panel.css.tsx
@@ -33,3 +33,11 @@ export const wordPanelContent = style({
 })
 
 export const audioContainer = style({ paddingLeft: "40%" })
+
+export const table = style({
+  width: "max-content",
+  borderBottom: "0px solid #ddd",
+  fontFamily: theme.fonts.body,
+  padding: "0px",
+  borderRadius: "0px",
+})
--- a/website/src/word-panel.css.tsx
+++ b/website/src/word-panel.css.tsx
@@ -85,11 +85,12 @@ export const audioContainer = style({ paddingLeft: "40%" })
 export const tableContainer = style({
   border: "0px solid transparent",
   width: "max-content",
+  margin: "0px",
 })
 
 export const tableCells = style({
-  border: "0px solid transparent",
-  borderRight: "5px solid transparent",
+  border: "4px solid transparent",
   fontFamily: theme.fonts.body,
-  padding: "3px",
+  padding: "0px",
+  wordWrap: "break-word",
 })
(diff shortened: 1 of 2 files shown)


loafofpiecrust:
Good first pass! Here's some thoughts off the cuff, I can check the code later. @patelShas Do you mind taking your first stab at code review on this one?
- Make sure to render the gloss for root morphemes. For example, in your screenshot `oohweel` should have a gloss as well.
- Add a small amount of margin between the table columns.
- Test with long morpheme descriptions and make sure it wraps properly on multiple lines.
-  Could you merge main into this branch so that we see what it looks like with collapsible sections? Lmk if you need help with that.

KushP10:
image

website/src/word-panel.tsx
```diff
@@ -129,6 +126,28 @@ export const WordPanel = (p: {
   }
 }
 
+export const PartLines = (p: {
+  segments: Dailp.FormFieldsFragment["segments"]
+}) => {
+  if (p.segments) {
```
loafofpiecrust, line 132: It's usually clearer to make the exceptional case the early return. In this case, that means:

```ts
if (!p.segments) {
  return null
}
```

And then the normal function body after that.

website/src/word-panel.tsx
```diff
@@ -129,6 +126,28 @@ export const WordPanel = (p: {
...
+    let length = p.segments.length
```
loafofpiecrust, line 133: Just trying to instill some better habits, let's call this `segmentCount` instead. And on line 136 instead of `part` lets call that `segment`, just to make all of the variable names consistent.

website/src/word-panel.tsx
```diff
@@ -129,6 +126,28 @@ export const WordPanel = (p: {
...
+    return (
+      <div>
+        {p.segments.map((part, index) => (
+          <tr>
+            <td className={css.table}>
```
loafofpiecrust, line 138: You called the style class `table` but it's for table cells. Could use a better name here.

website/src/word-panel.tsx
```diff
@@ -129,6 +126,28 @@ export const WordPanel = (p: {
...
+              {index > 0 ? "-" : null}
+              {part.morpheme}
+              {index === length - 1 ? null : "-"}
+            </td>
+            <td className={css.table}>{part.matchingTag?.learner?.title}</td>
```
loafofpiecrust, line 143: You can't just use `matchingTag.learner.title` because that's specifically the learner definition. Each tag set / abbreviation system has its own definition for each morpheme tag, so you need to use the one from the currently selected system. Check out the `morphemeDisplayTag` function in `website/src/types.ts` which will give you the right object given `matchingTag` as the first arg and the current tag set as the second arg.

website/src/word-panel.tsx
```diff
@@ -95,12 +95,9 @@ export const WordPanel = (p: {
...
           content={
             <>
-              <MorphemicSegmentation
-                segments={p.segment.segments}
-                onOpenDetails={p.onOpenDetails}
-                level={p.viewMode}
-                tagSet={p.tagSet}
-              />
+              <table className={css.table}>
+                <PartLines segments={p.segment.segments} />
```
loafofpiecrust, line 99: Can we call the component something more descriptive? For example, `VerticalMorphemicSegmentation` works. It's long but the name is more accurate. Also, move the `table` element inside that component and use it to replace the plain `div` you've got there now. General rule of thumb: if you have a `<div>` with no attributes and it's the only child of its parent element, then you don't need it at all.

website/src/word-panel.css.tsx
```diff
@@ -81,3 +81,11 @@ export const wordPanelHeader = style({
 })
 
 export const audioContainer = style({ paddingLeft: "40%" })
+
+export const table = style({
```
loafofpiecrust, line 85: I feel like the table itself and the table cells probably need different style classes.

KushP10:
added = following clitics but have to add = preceeding following segment

KushP10:
image

KushP10:
image

patelShas:
Something to nudge the gloss towards an earlier wrap around or set width might be nice. Most of the words end up looking like this, which is nice:
image

But then we also have words that look like this:
image

Jumping between them the word panel's size difference is jarring and sometimes changes the page enough that the selected word is no longer visible on the left and must be scrolled back to.

Additionally, while the gap between the words definition and parts at the bottom of the word parts panel is good for making clear their separation, I feel like its a bit much, as its wider than the gap between titles and content in the panels, and often closer to the top of Commentary than the rest of Word Parts.

loafofpiecrust:
Agreed on both counts. 1. I think different browsers might wrap table cells differently by default so just [explicitly tell them to wrap at word breaks](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Text/Wrapping_Text), and 2. For sure reduce the gap between the word parts and the overall word gloss.

loafofpiecrust (approved):

PR #155, opened 2022-03-29
added new cwkw theme

KushP10:
MicrosoftTeams-image (1)
MicrosoftTeams-image


loafofpiecrust (commented):
Good start, it needs a bit of refinement. Since we are dealing mainly with colors here, make sure to run contrast checks on a few pages. Firefox has a contrast check in the "Accessibility" tab, pretty sure Chrome has something similar. Chrome's lighthouse for sure does. Anything we ship to people needs to be WCAG AAA compliant on contrast.

loafofpiecrust (approved):

loafofpiecrust:
We can merge this since its a great start and then if theme changes come up later as we work on CWKW we'll add them in.

PR #156, opened 2022-03-31
[DP - 117] Preferences panel

patelShas:
Created a Preferences Panel accessible from a gear button in the header. 
image

Clicking on the gear will open a drawer where the options for preferences are laid out
image

The currently selected options have descriptions underneath that will update when the preferences are changed
image

Documents will appear as specified by preferences, which are saved across pages.
image

And will update when preferences are changed.
image


website/src/layout.tsx
```diff
@@ -9,13 +9,32 @@ import Footer from "./footer"
...
+import { HeaderPrefDrawer, selectedMode, selectedPhonetics } from "./mode"
 import { hideOnPrint, themeClass } from "./sprinkles.css"
+import { PhoneticRepresentation, ViewMode } from "./types"
 import "./wordpress.css"
 
 /* const ClientSignIn = lazy(() => import("./client/signin")) */
 
+// Set up experiencelevel
+export const preferencesContext = React.createContext({
+  expLevel: 0 as ViewMode,
```
loafofpiecrust, line 21: Use an enum value here instead of casting a number. I.e. `ViewMode.Story`.

website/src/layout.tsx
```diff
@@ -9,13 +9,32 @@ import Footer from "./footer"
...
+  expLevelUpdate: (p: ViewMode) => {},
+  phonRep: 0 as PhoneticRepresentation,
```
loafofpiecrust, line 23: There's no reason to use abbreviations here. Since we use the enum name ViewMode, we can call that variable viewMode and the other one phoneticRepresentation.

website/src/layout.tsx
```diff
@@ -9,13 +9,32 @@ import Footer from "./footer"
...
+  phonRepUpdate: (p: PhoneticRepresentation) => {},
```
loafofpiecrust, line 24: The common convention for state updates in React is `value` and `setValue` as opposed to what you've got here, `value` and `valueUpdate`

website/src/layout.tsx
```diff
@@ -9,13 +9,32 @@ import Footer from "./footer"
...
 import { MobileNav, NavMenu } from "./menu"
+import { HeaderPrefDrawer, selectedMode, selectedPhonetics } from "./mode"
 import { hideOnPrint, themeClass } from "./sprinkles.css"
+import { PhoneticRepresentation, ViewMode } from "./types"
 import "./wordpress.css"
 
 /* const ClientSignIn = lazy(() => import("./client/signin")) */
 
+// Set up experiencelevel
+export const preferencesContext = React.createContext({
```
loafofpiecrust, line 20: Move this into the types file or a new file and give it an upper camel case name. I.e. `PreferencesContext`. And then make a new component `PreferencesProvider` that wraps all the state management hooks you're doing in the Layout  omponent right now. Then, in Layout, you should only need to add `PreferencesProvider` wrapped around the return value.

website/src/layout.tsx
```diff
@@ -39,11 +58,17 @@ const Layout: React.FC = ({ children }) => {
                 Perseverance
               </span>
             </div>
+            <preferencesContext.Provider value={prefPack}>
```
loafofpiecrust, line 61: You definitely don't want to make multiple providers here, you should only have one. You can just wrap the whole return value of Layout in the provider.

website/src/mode.css.ts
```diff
@@ -23,3 +27,33 @@ export const levelLabel = sprinkles({
...
+      [mediaQueries.medium]: {
+        display: "block",
+      },
+    },
+  }
+])
+
+export const prefDrawer = style([
+  mobileWordPanel,
+  paddingX(hspace.edge), {
```
loafofpiecrust, line 56: Always run prettier before committing.

website/src/mode.tsx
```diff
@@ -123,38 +134,38 @@ export const ExperiencePicker = (p: { onSelect: (mode: ViewMode) => void }) => {
 export const PhoneticsPicker = (p: {
   onSelect: (phonetics: PhoneticRepresentation) => void
 }) => {
-  const radio = useRadioState({
-    state: selectedPhonetics(),
-  })
+  const [value, setValue] = useState(
+    selectedPhonetics() as PhoneticRepresentation
```
loafofpiecrust, line 138: Go ahead and move the `as ...` to the `selectedPhonetics` function so that you don't have to repeat it at every call site.

website/src/mode.tsx
```diff
@@ -123,38 +134,38 @@ export const ExperiencePicker = (p: { onSelect: (mode: ViewMode) => void }) => {
...
   useEffect(() => {
-    Cookies.set("phonetics", radio.state!.toString(), {
+    Cookies.set("phonetics", value.toString(), {
       sameSite: "strict",
       secure: true,
     })
-    p.onSelect(radio.state as PhoneticRepresentation)
-  }, [radio.state])
-
+    p.onSelect(value as PhoneticRepresentation)
```
loafofpiecrust, line 147: I don't think you need  the `as` cast here

website/src/mode.tsx
```diff
@@ -225,3 +236,43 @@ const TagSetOption = (p: { radio: RadioStateReturn; level: string }) => {
     </Tooltip>
   )
 }
+
+export const PrefPanel = () => {
+  const preferences = useContext(preferencesContext)
+  return (
+    <div>
+      <h3>Display Mode:&ensp; </h3>
+      {<ExperiencePicker onSelect={preferences.expLevelUpdate} />}
```
loafofpiecrust, line 245: I'm not sure why you have `ensp` and a space here. Also, I don't think you need headings here, normal `label` elements should work just fine. And then instead of ensp you just need a normal space. To write a space between elements on different lines, you can put `{" "}` on the first line to make an explicit space.

website/src/mode.tsx
```diff
@@ -225,3 +236,43 @@ const TagSetOption = (p: { radio: RadioStateReturn; level: string }) => {
...
+      <p>{levelNameMapping[preferences.expLevel].details}</p>
+      <h3>Romanization Method:&ensp; </h3>
+      {<PhoneticsPicker onSelect={preferences.phonRepUpdate} />}
+      <p>{phoneticRepresentationMapping[preferences.phonRep].details}</p>
```
loafofpiecrust, line 249: Could you look into how we might be able to semantically connect the picker to this description? Maybe `aria-describedby` would do the trick.

website/src/mode.tsx
```diff
@@ -225,3 +236,43 @@ const TagSetOption = (p: { radio: RadioStateReturn; level: string }) => {
...
+        {...dialog}
+        aria-label="Open Preferences Drawer"
+        className={css.prefButton.button}
+      >
+        <MdSettings size={32} />
+      </DialogDisclosure>
+      <DialogBackdrop {...dialog} className={css.prefBG}>
+        <Dialog
+          {...dialog}
+          as="nav"
```
loafofpiecrust, line 269: This isn't a navigation element so we shouldn't use `nav`. Either `div` or perhaps `aside` should work here.

website/src/pages/documents/document.page.tsx
```diff
@@ -163,12 +175,9 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
     setCurrContents: selectAndShowWord,
   }
 
-  const [phoneticRepresentation, _setPhoneticRepresentation] =
-    useState<PhoneticRepresentation>(selectedPhonetics())
+  const phoneticRepresentation = useContext(preferencesContext).phonRep
```
loafofpiecrust, line 178: Make a hook inside whatever file you make the `PreferencesProvider` component, a new hook called `usePreferences` that wraps this useContext call. Generally, convention is that you never export a context directly, you only use it from abstractions like a custom provider component or custom hook. Also, you only need to call the hook once here, like `const { viewMode, phoneticRepresentation } = usePreferences()`.

website/src/segment.tsx
```diff
@@ -116,9 +130,8 @@ export const AnnotatedForm = (
         </div>
         {p.segment.simplePhonetics ? (
```
loafofpiecrust, line 131: Make sure to change these render conditions since you aren't using `simplePhonetics` directly in rendering now.

website/src/mode.tsx
patelShas: The label elements look a lot like regular text, even similar to the regular text. I'm not sure if this is really what we want?

image

website/src/mode.tsx
patelShas: Do you mean just showing better how the desccriptions correspond to the selectors within the code? Or are there functional changes as well. I don't really understand what it means to semantically connect them.

patelShas:
Most changes have been made, and the page and preferences panel look largely the same, though I'm still unsure about keeping the labels over the preferences options as `<label>`'s since they look like this:
image
So some feedback would be nice. 

Aside from that I think everything's good.

loafofpiecrust:
Awesome, thanks. For the labels, I think if we add styling to the labels then using that element I think should work just fine. Not sure they need much more than bolding.

In terms of the panel being a slide-out, I'm not sold on it, at least on desktop. I feel like we shouldn't be dimming the contents while you change settings so that you can properly preview them. Perhaps just a hovering disclosure would work well? I'll sketch something in a bit to show you what I mean.

patelShas:
Rather than a side drawer, the preferences panel is now appears in the center. Also added a "Ducoments Settings" header and a close button.

This is how it looks on desktop:
image

And this is how it looks on mobile:
image

loafofpiecrust:
Looking great! Last things on this before we merge it: let's make the label always sit above the dropdowns, instead of forming a row. And then let's make the labels bold.

loafofpiecrust (approved):

PR #167, opened 2022-04-15
[DP-101] Added Table of Contents functionality

KushP10:
image
Sample image

image
Added a new type of data to allow for recursive functionality of subsection


loafofpiecrust:
So you still need to handle:
1. Introduction chapters. If you look at the wireframe image, there are intro chapters with roman numerals before the normally numbered chapters and we should include that. So in your data structure, that could look like two fields `introChapters` and `chapters` with each the same type.
2. This may already work fine, but just make sure to include 10 or more chapters at some hierarchy level to make sure that chapter titles are all aligned with each other.

patelShas:
I think it looks pretty good, though I have one suggestion. Right now the distinction between intoduction and body nodes is only made at the top level of TOC, so in any recursive cases, all children of a node are assumed to be body nodes. While this would work most of the time, I can see a case where in the future we might want to have an arrangement kinda like:

- intro 1 
- intro 2
- content 1
  - intro for content 1
  - content 1 part 1
  - content 1 part 2

So for this I think it would be helpful to move the distinction between intro and body nodes into TOCData rather than TOC, perhaps by either making the children property into two arrays or by splitting into two properties.

loafofpiecrust:
@patelShas Good idea, but I think we can hold off on that until we have actual data and then we'll give the data structure a harder look. Approving this.

loafofpiecrust (approved):

PR #175, opened 2022-04-29
[DP-99 & 100] A Layout component for CWKW

patelShas:
A page has been created showing off what the CWKW layout looks like on the url [site]/cwkw-placeholder.

image

A new menu was created for this layout, and the menu components have also been changed to accept menuID's rather than being hard-coded for a specific one. So, the menu component still works for the main page:

image

Creating the cwkw layout also required making the preferences and mobile menu buttons change colors so they would remain visible. To show them, here is the site on mobile:

image


website/src/cwkw/cwkw-layout.css.ts
```diff
@@ -0,0 +1,18 @@
+import { style } from "@vanilla-extract/css"
+import * as baseLayout from "../layout.css"
+import { colors } from "./theme.css"
+
+export const header = style([baseLayout.header])
```
loafofpiecrust, line 5: If the style is just reusing another one, I think you can just do `export const header = baseLayout.header`. You could reduce this even further by doing `export { header, headerContents, siteTitle, siteLink } from "../layout.css"`

website/src/graphql/wordpress/queries.graphql
```diff
@@ -34,3 +34,23 @@ query MainMenu {
     }
   }
 }
+
+query MenuByID($slug: Int!) {
+  menus(where: { id: $slug }) {
+    nodes {
+      databaseId
+      menuItems(where: { parentDatabaseId: 0 }) {
```
loafofpiecrust, line 42: Hmm why are you using `parentDatabaseId: 0` here? Also, I would rather us have one query definition for both menus instead of a separate one for the main site menu.

website/src/menu.css.ts
```diff
@@ -53,7 +61,7 @@ export const drawerItem = style([
     display: "inline-block",
     padding: vspace.quarter,
     textDecoration: "none",
-    color: colors.text,
+    color: "black",
```
loafofpiecrust, line 64: Why this change? I'd rather add another color to the theme that handles this instead of using a hard coded color.

website/src/menu.css.ts
```diff
@@ -45,6 +46,13 @@ export const navLink = style([
   },
 ])
 
+export const subMenuItems = style([
+  navLink,
+  {
+    color: "black",
```
loafofpiecrust, line 52: Use a theme color, I'd like to avoid hard-coded colors where possible.

website/src/menu.css.ts
```diff
@@ -10,6 +10,7 @@ import sprinkles, {
   vspace,
 } from "src/sprinkles.css"
 import { paddingX, paddingY } from "src/style-utils"
+import * as cwkwTheme from "./cwkw/theme.css"
```
loafofpiecrust, line 13: Never use themes directly, always use the variables from the theme contract defined in the `sprinkles.css` file. That way if you switch themes, the styles will update accordingly.

website/src/menu.css.ts
```diff
@@ -73,6 +81,11 @@ export const navButton = style([
   },
 ])
 
+export const navButtonChoices = styleVariants({
+  cwkw: [navButton, { color: cwkwTheme.colors.text }],
```
loafofpiecrust, line 85: So here, I wouldn't make variants. Instead change the `navButton` styles so that switching themes does this for us.

website/src/menu.tsx
```diff
@@ -16,20 +16,28 @@ import {
...
   navLink,
   navMenu,
+  subMenuItems,
 } from "./menu.css"
 import { closeBlock } from "./sprinkles.css"
 
-export const NavMenu = () => {
+export const NavMenu = (p: { menuID: number }) => {
   const location = useLocation()
-  const [{ data }] = Wordpress.useMainMenuQuery()
```
loafofpiecrust, line 27: Since you swapped out the main menu query, can we just get rid of it then?

website/src/menu.tsx
```diff
@@ -16,20 +16,28 @@ import {
...
-  const menuItems = data?.menuItems?.nodes
+  const [{ data }] = Wordpress.useMenuByIdQuery({
+    variables: { slug: p.menuID },
+  })
+  const menus = data?.menus?.nodes
```
loafofpiecrust, line 32: Why did you need to query `menus` instead of `menuItems`?

website/src/menu.tsx
```diff
@@ -100,11 +108,18 @@ const SubMenu = ({ item, location }: { location: Location; item: any }) => {
   )
 }
 
-export const MobileNav = () => {
+export const MobileNav = (p: { menuID: number; cwkw?: boolean }) => {
```
loafofpiecrust, line 111: Yeah so get rid of this `cwkw` prop and make all necessary changes through the theme.

website/src/mode.css.ts
```diff
@@ -46,7 +47,10 @@ export const prefButtonShell = style({
   flex: 1,
 })
 
-export const prefButton = collPanelButton
+export const prefButton = styleVariants({
+  cwkw: [collPanelButton, { color: cwkwTheme.colors.text }],
+  base: [collPanelButton],
+})
```
loafofpiecrust, line 53: Same thing here, don't make a cwkw variant, use the theme for this change.

website/src/pages/documents/document.page.tsx
```diff
@@ -10,6 +10,7 @@ import { DocumentAudio } from "src/audio-player"
 import { Breadcrumbs } from "src/breadcrumbs"
 import { Button } from "src/components"
 import Link from "src/components/link"
+import CWKWLayout from "src/cwkw/cwkw-layout"
```
loafofpiecrust, line 13: This looks like an unnecessary import

website/src/menu.tsx
```diff
@@ -16,20 +16,28 @@ import {
...
 } from "./menu.css"
 import { closeBlock } from "./sprinkles.css"
 
-export const NavMenu = () => {
+export const NavMenu = (p: { menuID: number }) => {
   const location = useLocation()
-  const [{ data }] = Wordpress.useMainMenuQuery()
-  const menuItems = data?.menuItems?.nodes
+  const [{ data }] = Wordpress.useMenuByIdQuery({
+    variables: { slug: p.menuID },
```
loafofpiecrust, line 30: I think we can call the query variable `id` instead of `slug` because its not a slug right?

website/src/menu.tsx
patelShas: Querying menuItems seemed to just return items without regard to which menu they came from, and their parentID was usually null and so couldn't be filtered by that. Mean while querying menus made it easy to see all the nodes of a specific wordpress menu by filtering by ID. So, it seemed easier to query menus than menuItems.

website/src/menu.css.ts
patelShas: There was an odd quirk that made it so menu items were not visible against the background, which has been fixed.

website/src/graphql/wordpress/queries.graphql
patelShas: I don't know why but including it makes the graphql query return Contact-Us, but excluding it means Contact-Us does not appear. I can't tell why, as the parentDatabaseID and parentID themselves are not different from other nodes. So, including this (where: { parentDatabaseId: 0 }) was the simplest way to bring back Contact-Us.

patelShas:
Made most changes as well as some changes to the cwkw theme so that elements would be more visible.

image

website/src/cwkw/theme.css.ts
```diff
@@ -28,4 +27,10 @@ export const themeClass = createTheme(theme, {
...
     smallCaps: "Charis SIL",
   },
-})
\ No newline at end of file
+})
+
+export const colors = {
+  ...theme.colors,
+  transparent: "transparent",
+  inherit: "inherit",
```
loafofpiecrust, line 35: You don't need to define these in the colors object because they'll never change. When I said use only theme colors, I should've said "unless they're special values". For example we'll never need to change what `inherit` means but we may want to change what "the text color" means.

website/src/menu.css.ts
loafofpiecrust: This still needs to be using the `colors` object.

website/src/pages/cwkw-placeholder.page.tsx
```diff
@@ -0,0 +1,16 @@
+import CWKWLayout from "src/cwkw/cwkw-layout"
+
+const CWKWPlaceholderPage = () => (
+  <CWKWLayout>
+    <main>
```
loafofpiecrust, line 5: Could you add here:

```tsx
<Helmet>
  <meta name=”robots” content=”noindex,nofollow”/>
</Helmet>
```

Helmet comes from `react-helmet`, you can look for other usages to see examples. This will prevent the page from being indexed since we don't want anyone to get linked to it.

website/src/sprinkles.css.ts
```diff
@@ -41,7 +41,7 @@ export const themeClass = createTheme(theme, {
   },
   colors: {
     header: "#f7eeed",
-    button: "#f7eeed",
+    button: "black",
```
loafofpiecrust, line 44: Why are you changing a color in the base theme?

website/src/cwkw/theme.css.ts
patelShas: Yeah, I've realized since that this whole css declaration of themes is unnecessary and I just misunderstood how the themes work when adding it in.

website/src/sprinkles.css.ts
patelShas: Beforehand the button was the same color as the header, so making buttons and links have the button color meant they were invisible in the base theme. So to fix this, I tried changing this to black. I see though this doesn't work well with other elements that already used the button color.

loafofpiecrust (approved):

PR #179, opened 2022-05-17
[DP-149] Added wordpress reference usability to parser options

KushP10:


.sqlfluff 2
```diff
@@ -0,0 +1,32 @@
+[sqlfluff]
```
loafofpiecrust, line 1: Why is this a new file in your branch? Maybe your local main branch that you based this branch on is several commits behind?

website/src/wordpress-page.tsx
```diff
@@ -4,6 +4,11 @@ import parse, {
   attributesToProps,
   domToReact,
 } from "html-react-parser"
+import {
+  NodeWithChildren,
+  isText,
+} from "html-react-parser/node_modules/domhandler"
```
loafofpiecrust, line 10: What's up with this import? Never import from a dependency's node_modules. Does `html-react-parser` re-export these items so you can import them from there?

website/src/wordpress-page.tsx
```diff
@@ -41,6 +46,12 @@ export const WordpressContents = ({ content }: { content: string }) => {
 
 const parseOptions: HTMLReactParserOptions = {
   replace(node) {
+    if (isText(node) && node.data.startsWith("[") && node.data.endsWith("]") && node.data.indexOf(":") !== -1) {
+      const text = node.data.split(/[\[,\],:,-]/).filter((x) => x !== "") // figure out why empty strings in start and end of array
```
loafofpiecrust, line 50: How about instead of using `split`, you use capture groups in the regex? That should be more robust and follow a more explicit syntax. With your current code, this footnote-like text would pass your checks: `[that book]: found at the library--with dust on top` and turn into `["that book", "found at the library", "", "with dust on top"]`. Which means it's not strict enough. What you're reaching for is predefining a regex pattern for the whole reference, then checking if the node text matches the pattern, then if so using the matching results. [Here's a doc on captures in JS regex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges)

website/src/wordpress-page.tsx
```diff
@@ -56,7 +67,21 @@ const parseOptions: HTMLReactParserOptions = {
...
     return undefined
   },
 }
+function pullWords(
+  arg0: string | undefined,
+  arg1: string | undefined,
+  arg2?: string | undefined
+): false | void | object | JSX.Element | null | undefined {
+  throw new Error("Function not implemented.")
+}
```
loafofpiecrust, line 87: What's up with the typing on this? I'd say just define a very simple React component which renders the parts of the reference to demonstrate that they're properly captured. Say, transforming `[WJ20:1-5]` into `Words #1 through #5 in WJ20` or something, until we hook up the actual content. But we don't want to throw an error as our stop-gap.

website/.envrc 2
```diff
@@ -0,0 +1,2 @@
+source_up
+layout node
```
loafofpiecrust, line 2: Still not sure why this file is in this PR

loafofpiecrust (approved):
Generally looks good to me, should set us up for the future work. I'm still wondering about a short doc outlining the requirements on the WordPress/content side?

loafofpiecrust (approved):

PR #180, opened 2022-05-17
[DP - 146] initial database structure for the edited collections

patelShas:
Created some new tables/queries in sql to add support for edited collections. 

Namely, added a "chapter" table that will allow chapters in edited collections to be represented, some sample data looking like:
image

The ltree field is useful and another query allows groups of chapter to be selected for based on how we want to organize them, which will be helpful, For example:
image
For intro chapters. Similarly, different Ltrees can be used for different collections or subcollections.

### DIFF
--- a/types/migrations/20220205042039_init.sql
+++ b/types/migrations/20220205042039_init.sql
@@ -3,8 +3,6 @@
 
 create extension if not exists "uuid-ossp";
 
-create extension if not exists "ltree";
-
 create domain autouuid uuid default uuid_generate_v4 ();
 
 -- Base website: Language > Group > Document
@@ -37,16 +35,6 @@ create table document (
   is_reference boolean not null,
   written_at date,
   audio_slice_id uuid references media_slice (id) on delete set null
-);
-
-create table chapter (
-  id autouuid primary key,
-  chapter_title text not null,
-  document_title text,
-  author text,
-  document_id text,
-  wordpress_id bigint,
-  collection_path ltree not null,
 );
 
 create table iiif_source (
--- a/types/migrations/20220205042039_init.sql
+++ b/types/migrations/20220205042039_init.sql
@@ -42,12 +42,12 @@ create table document (
 create table chapter (
   id autouuid primary key,
   chapter_title text not null,
-  document_title text not null,
-  author text not null,
-  document_id bigint,
+  document_title text,
+  author text,
+  document_id text,
   wordpress_id bigint,
   collection_path ltree not null,
-)
+);
 
 create table iiif_source (
   id autouuid primary key,
--- a/types/migrations/20220205042039_init.sql
+++ b/types/migrations/20220205042039_init.sql
@@ -3,6 +3,8 @@
 
 create extension if not exists "uuid-ossp";
 
+create extension if not exists "ltree";
+
 create domain autouuid uuid default uuid_generate_v4 ();
 
 -- Base website: Language > Group > Document
@@ -36,6 +38,16 @@ create table document (
   written_at date,
   audio_slice_id uuid references media_slice (id) on delete set null
 );
+
+create table chapter (
+  id autouuid primary key,
+  chapter_title text not null,
+  document_title text not null,
+  author text not null,
+  document_id bigint,
+  wordpress_id bigint,
+  collection_path ltree not null,
+)
 
 create table iiif_source (
   id autouuid primary key,
--- a/types/migrations/20220518174701_edited_collections.sql
+++ b/types/migrations/20220518174701_edited_collections.sql
@@ -16,7 +16,7 @@ create table chapter (
   title text not null,
   document_id uuid references document(id),
   wordpress_id bigint,
-  index_tree ltree not null,
+  index_tree ltree not null unique,
   slug text not null
 );
 
--- a/types/migrations/20220518174701_edited_collections.sql
+++ b/types/migrations/20220518174701_edited_collections.sql
@@ -7,8 +7,8 @@ create extension if not exists "ltree";
 create table edited_collection (
   id autouuid primary key,
   title text not null,
-  wordpress_menu_id bigint;
-  slug ltree not null unique -- keep it the same type as chapter ltree
+  wordpress_menu_id bigint,
+  slug text not null unique -- keep it the same type as chapter ltree
 );
 
 create table collection_chapter (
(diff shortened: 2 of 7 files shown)


types/migrations/20220205042039_init.sql
```diff
@@ -3,6 +3,8 @@
 
 create extension if not exists "uuid-ossp";
 
+create extension if not exists "ltree";
```
loafofpiecrust, line 6: Never ever edit database migrations that are already merged to main and applied on live deployments. We always have to make new migration files for new changes, because postgres skips completed migrations based on the file name and our migration process will fail if an existing migration is edited. If you navigate to the `types` folder, make a new migration with `sqlx migrate new migration_name`. It'll automatically prepend the date and create a .sql file for you.

types/migrations/20220205042039_init.sql
```diff
@@ -37,6 +39,16 @@ create table document (
   audio_slice_id uuid references media_slice (id) on delete set null
 );
 
+create table chapter (
+  id autouuid primary key,
+  chapter_title text not null,
+  document_title text,
+  author text,
+  document_id text,
```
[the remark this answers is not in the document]
loafofpiecrust: Don't worry too much about the insert for now, we'll have to tweak that when we write the code for data ingestion anyway. FYI tho The variable placeholder will still work with a `uuid`, yes. Another option is to pass in document short name like `WJ20` as $1 and then instead of doing `values ($1, ...)` you can do `select document.id, $2, $3 from document where document.short_name = $1` or similar.

types/queries/insert_chapter.sql
```diff
@@ -0,0 +1,4 @@
+insert into chapter (chapter_title, document_title, author, document_id, wordpress_id, collection_path)
+values ($1, $2, $3, $4, $5, $6)
+on conflict (chapter_title, document_title)
```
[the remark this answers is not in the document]
loafofpiecrust: Yeah, the only fields that should be unique are `id` (is by nature of being a primary key) and probably the path ltree (we don't want two chapters trying to have the same URL).

types/migrations/20220205042039_init.sql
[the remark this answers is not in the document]
loafofpiecrust: If you check out the [examples here](https://www.postgresql.org/docs/current/sql-insert.html) there's some `insert into ... select ...` examples.

loafofpiecrust:
For the collection table, these are the fields I think we need to start with:
- `id: uuid`
- `title: text`
- `slug: text` (unique)

Also by the way, the reason we need a `uuid` as primary key for everything even if they have a unique `slug` (as opposed to using `slug` as the primary key) is that internal database references are more reliable if the primary key never changes, plus its easy to remember that *all* primary/foreign keys should be uuid. We can have other entities point to a specific collection and never need to update that reference even if the collection slug changes. In this context, `id` is called a "surrogate key" and `slug` is a "natural key".

patelShas:
Fleshed things out more, including the addition of a collections table and chapter_attribution table for authors.
Biggest change to the chapter table is instead of having an ltree of slugs and an index for position, it has an ltree of indexes and a slug for itself:
image
The index tree was chosen because it allows the same hierarchy to be visible while also making ordering easier. It also includes the slug of the parent collection.

The two big sample queries are

- update_chapter_place, which will let you move a chapter and its children to a new parent, and/or change what index it has, and
- collection_contents, which will return the information on all chapters in a specified collection, ordered by their indexes in each parent.

types/migrations/20220518174701_edited_collections.sql
```diff
@@ -0,0 +1,28 @@
+-- Add migration script here
+
+-- A file for collections and chapters
+
+create extension if not exists "ltree";
+
+create table collections (
```
loafofpiecrust, line 7: Our convention is to use singular table names, so `collection`

types/migrations/20220518174701_edited_collections.sql
```diff
@@ -0,0 +1,28 @@
...
+  id autouuid primary key,
+  collection_name text not null,
+  collection_desc text,
+  slug ltree not null unique -- keep it the same type as chapter ltree
```
loafofpiecrust, line 11: Collections won't ever be nested within each other, so the slug should just be `text`

types/migrations/20220518174701_edited_collections.sql
```diff
@@ -0,0 +1,28 @@
+-- Add migration script here
+
+-- A file for collections and chapters
+
+create extension if not exists "ltree";
+
+create table collections (
+  id autouuid primary key,
+  collection_name text not null,
+  collection_desc text,
```
loafofpiecrust, line 10: Hmm we don't need the prefixes here, I'd go for `title` and I'm not positive that we need description text yet because that might be Wordpress content.

types/migrations/20220518174701_edited_collections.sql
```diff
@@ -0,0 +1,28 @@
...
+  slug ltree not null unique -- keep it the same type as chapter ltree
+);
+
+create table chapter (
```
loafofpiecrust, line 14: Let's call this `collection_chapter` to indicate the relationship.

types/migrations/20220518174701_edited_collections.sql
```diff
@@ -0,0 +1,28 @@
...
+  id autouuid primary key,
+  title text not null,
+  document_id uuid references document(id),
+  wordpress_id bigint,
+  index_tree ltree not null unique,
+  slug text not null
+);
+
+create table chapter_attribution (
```
loafofpiecrust, line 23: If we call the above `collection_chapter`, then let's call this `collection_chapter_attribution`

types/migrations/20220518174701_edited_collections.sql
patelShas: `collection` seems to be recognized as a data type or something known to sql, so `edited_collection` would be a good choice rather than `collection` to avoid confusion. Will be made singular though.

patelShas: `collection` gets highlighted as a recognized term, and its apparently a data type in sql. To avoid confusion/risk, I think `edited_collection` is a good name.

loafofpiecrust (commented):
Thought I submitted these comments, but we need to wrap up this PR

PR #182, opened 2022-05-26
added preloading to word's audio player

patelShas:
The word panel's have been told to preload audio. Based on my understanding this means the audio panel should have loaded the audio as the page was loaded. Previously this mean there was some time between clicking "play" and hearing audio in which seemingly nothing would happen. Now this doesn't happen.


website/src/pages/documents/document.css.ts
```diff
@@ -182,6 +182,7 @@ export const contentSection2 = style([
   sprinkles({ display: { any: "none", medium: "block" } }),
   {
     flex: 1,
+    width: "25vw",
```
loafofpiecrust, line 185: We shouldn't use `vw` units unless totally necessary

loafofpiecrust:
As we've talked about, I think this PR could include changing the play/pause icon to something else for the loading state and also disable that button from being clicked while the audio is still loading, to avoid some of our issues there.

loafofpiecrust:
Just looking at the functional audio player component again, why is `howl` in a `useState`? It looks like we never call `setHowl` at all so I think we can actually use `useMemo` instead.

patelShas:
Implemented suggested changes. Instead of the button icon being a react state, now there is a state isPlaying. The useEffect was removed, and now the most frequently running one only changes for slices, essentially just for when a new word is selected. Howl was also moved from a useState to a useMemo.
In terms of features, 

- Preloading is added
- if the audio is loading, the icon should reflect that
- You can pause and play a word's audio.
- Even if you've paused the audio, moving to a new word will play the new word's audio normally.
- The button reflects the playing status too.

loafofpiecrust:
If I play audio for one word, pause halfway through, then select a different word, the progress bar stays in the same place. Seems like we don't want this, make sure that the progress state gets reset when doing this.

website/src/audio-player.tsx
```diff
@@ -45,201 +54,93 @@ interface Props {
...
-    this.nextStep = this.nextStep.bind(this)
-    this.onPlay = this.onPlay.bind(this)
-    this.onSeek = this.onSeek.bind(this)
-  }
 
-  onPlay() {
-    requestAnimationFrame(this.nextStep)
-  }
+const AudioComponent = (props: Props) => {
+  const audio = useRef(new Audio(props.audioUrl))
```
loafofpiecrust, line 60: Remember that this function runs fully on every render, so you're calling `new Audio` on every render. I would move this to a `useMemo` or similar because I don't think you ever change the value.

website/src/audio-player.tsx
```diff
@@ -45,201 +54,93 @@ interface Props {
...
-
-  if (props.slices) {
-    Object.assign(howl, {
-      _sprite: {
-        __default: [props.slices.start, props.slices.end - props.slices.start],
-        sound: [props.slices.start, props.slices.end - props.slices.start],
-      },
-    })
-  }
+  }, [progress])
```
loafofpiecrust, line 102: This is missing a dependency on end or props.slices

website/src/audio-player.tsx
```diff
@@ -45,201 +54,93 @@ interface Props {
...
-    }))
-    if (this.state.howl.playing()) {
-      requestAnimationFrame(this.nextStep)
+  const [progress, setProgress] = useState(0)
+  const [loadStatus, setLoadStatus] = useState(0)
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setProgress(audio.current.currentTime)
+      setLoadStatus(audio.current.readyState)
```
loafofpiecrust, line 67: There's no listeners for load progress at all on the Audio api?

patelShas: There are some event listeners and loading can be rearranged on that. There is also an event listener for the current time, but I think to avoid that since using that onEvent winds up being much choppier than checking current time every set time period.

patelShas:
I've added in suggested changes. I'm not quite sure why the checks have all failed, since on my local copy it still runs fine. But yeah,

- switched from howl to audio
- single component for all audio players
- loading icon when there isn't enough data to play through

PR #184, opened 2022-06-13
[DP - 153] Mixed a scrolling bug on mobile word panels

patelShas:
Fixed a bug in the word panels where if content on mobile overflowed, you couldn't scroll the word panel. Now, if you are on mobile and content overflows, you can scroll on either main content or word details depending on where you scroll from.

Also, restricted word panel width on desktop so it wouldn't exceed a certain amount. I thought that had been done before, but apparently not.

### DIFF
--- a/website/src/pages/documents/document.css.ts
+++ b/website/src/pages/documents/document.css.ts
@@ -182,7 +182,7 @@ export const contentSection2 = style([
   sprinkles({ display: { any: "none", medium: "block" } }),
   {
     flex: 1,
-    maxWidth: "33vw",
+    maxWidth: "20rem",
   },
 ])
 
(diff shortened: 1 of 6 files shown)


loafofpiecrust:
I think what we want is just to prevent scrolling the main content while you have the modal open on mobile.

website/src/pages/documents/document.css.ts
```diff
@@ -182,6 +182,7 @@ export const contentSection2 = style([
   sprinkles({ display: { any: "none", medium: "block" } }),
   {
     flex: 1,
+    maxWidth: "33vw",
```
loafofpiecrust, line 185: Using `vw` units doesn't really make sense for desktop (this is for desktop right?), because the content is at a fixed width. So this would make more sense as a `px` value or `rem` value perhaps.

loafofpiecrust (approved):

PR #185, opened 2022-06-15
Add manuscript validation step to the beginning of migrations

loafofpiecrust:
- Collects all errors from spreadsheets and prints them all together at the end of the log, if there's one or more failures.
- Limitation: only shows the google sheet ID in each error message, because it failed to fully process the sheet into a document struct.

### DIFF
--- a/migration/src/main.rs
+++ b/migration/src/main.rs
@@ -23,6 +23,9 @@ async fn main() -> Result<()> {
     dotenv::dotenv().ok();
     pretty_env_logger::init();
 
+    println!("Validating manuscript spreadsheets...");
+    validate_documents().await?;
+
     let db = Database::connect().await?;
 
     println!("Migrating Image Sources...");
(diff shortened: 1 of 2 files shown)


GracefulLemming (approved):
I think this looks good overall! Im wondering if we can provide some info about each error (in addition to "Failed to process") to help nail down issues in a time efficient manner. I think that detail would allow translation and audio teams to quickly diagnose and resolve issues during migration. Otherwise, from the perspective of the translation and audio teams, I think  its a really nice feature to have all of the errors display at once as you have it (even if it uses the sheet id)!

loafofpiecrust:
@GracefulLemming Yeah so it'll only print that plain failed error with no details if it failed to parse the metadata sheet, since we allow that step to fail right now. For other issues it'll print the underlying error too.

loafofpiecrust:
I'm gonna merge this for now then and we can try to make this better as needed.

PR #186, opened 2022-06-16
[DP-60] Create custom login UI using AWS cognito library

chanmioh:
CHANGES:
- Added a login button at the top right of the website; clicking this will bring the user to the login page at `/login`
- On the login page, entering login credentials and submitting login will attempt a user authentication. If login is successful, the React context holding info about the user is updated accordingly. 
- With this context, the user's JWT token can be accessed + other authentication info if needed.
- If the user is logged in, the login button of the website will function as a Logout.

- Login UI: image
- What user sees while logged in: image


website/src/auth.tsx
```diff
@@ -2,52 +2,71 @@ import { HubCallback } from "@aws-amplify/core"
 import Amplify, { Auth, Hub } from "aws-amplify"
 import React, { createContext, useContext, useState } from "react"
 
-const UserContext = createContext<CognitoUser | null>(null)
+import {
+  CognitoUser,
+} from 'amazon-cognito-identity-js'
```
loafofpiecrust, line 7: Right now, I think this is a transitive dependency from the amplify library. Could you run `yarn add amazon-cognito-identity-js`?

website/src/auth.tsx
```diff
@@ -2,52 +2,71 @@ import { HubCallback } from "@aws-amplify/core"
...
+
+type UserContextType = {
+  user: CognitoUser | null;
+  setUser: (value: CognitoUser | null) => void;
+};
+
+export const UserContext = createContext<UserContextType>({} as UserContextType);
```
loafofpiecrust, line 14: General convention is not to export context objects directly (only `XProvider` and `useX`). So I'd say either all the code that needs to call `setUser` goes into one file, or exposing `setUser` via the hook (or a new hook). I lean toward the former because generally new code beyond this shouldn't need to call `setUser`.

website/src/client/layout.tsx
```diff
@@ -1,16 +1,17 @@
...
-import { UserProvider, useCredentials } from "../auth"
+import { useCredentials } from "../auth"
+import { Provider as GraphQLProvider } from "urql"
 
-const LayoutClient = (p: { children: any }) => {
-  const creds = useCredentials()
-  const token = creds?.signInUserSession.idToken.jwtToken ?? null
+export const LayoutClient = (p: { children: any }) => {
+  // const creds = useCredentials()
+  const token = useCredentials() ?? null; // uses credentials
```
loafofpiecrust, line 8: I would move this `?? null` to inside `useCredentials` so that callsites don't need it.

website/src/layout.tsx
```diff
@@ -18,31 +20,34 @@ import "./wordpress.css"
...
-          <div className={css.contentContainer}>
-            <h1 className={css.siteTitle}>
-              <Link className={css.siteLink} href="/">
-                DAILP
-              </Link>
-            </h1>
-            <span className={css.subHeader}>
-              Digital Archive of Indigenous Language Persistence
-            </span>
+      <LayoutClient>
```
loafofpiecrust, line 23: Hmm I do think we need `LayoutClient` at this scope but let's move the `Helmet` usage before it since that doesn't call graphql.

```html
<Helmet ...>
</Helmet>

<LayoutClient>
...
```

website/src/auth.tsx
```diff
@@ -2,52 +2,71 @@ import { HubCallback } from "@aws-amplify/core"
...
-          setUser(null)
-          break
-      }
-    }
-    Hub.listen("auth", listener)
-    return () => Hub.remove("auth", listener)
-  }, [])
+  const [user, setUser] = useState<CognitoUser | null>(null);
+
+  // React.useEffect(() => {
```
loafofpiecrust, line 19: We definitely need something here to restore remembered logins, the cognito library docs have an example of it (use case 16).

website/src/client/signin.tsx
```diff
@@ -2,19 +2,19 @@ import { Auth } from "aws-amplify"
...
 
-const ClientSignIn = () => {
-  const creds = useCredentials()
-  if (creds) {
-    return <Button onClick={() => Auth.signOut()}>Sign Out</Button>
-  } else {
-    return <Link href="/signin">Sign In</Link>
-  }
-}
+// const ClientSignIn = () => {
```
loafofpiecrust, line 7: If you think these components are not needed anymore, you can delete the file.

website/src/pages/login.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <div className={loginHeader}>
+        {/* show a logout button if user is signed in, otherwise show login */}
+        {context.user !== null ?
+          <Link href=""
+            onClick={(e) => {
+              e.preventDefault();
+              context.user?.signOut();
+              context.setUser(null);
+            }}>Logout</Link>
+          : <Link href="/login">Login</Link>}
```
loafofpiecrust, line 128: Let's do "Log out" and "Log in" (two words)

website/src/pages/signup.page.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from "react"
+import Link from "src/components/link";
+import Layout from "src/layout"
+import { centeredColumn } from "src/sprinkles.css"
+import { centeredForm, loginFormBox, positionButton, skinnyWidth, submitButton } from "./login.css";
+
+// TODO: this page is  incomplete
+const SignupPage = () => {
```
loafofpiecrust, line 8: FYI you and I hadn't talked about this together yet but we are planning on having more like an "Apply to join" flow, but that might manifest as allowing registration for a very limited privilege account until it's approved to contribute. So the registration page will need further thought but for now let's use `register` as the path to avoid "Sign in/out/up" entirely.

website/src/pages/reset-password.page.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { centeredColumn } from "src/sprinkles.css"
+import { centeredForm, loginFormBox, positionButton, skinnyWidth, submitButton } from "./login.css";
+
+// TODO: this page is incomplete
+const ResetPasswordPage = () => {
+
+    const [email, setEmail] = useState("");
+
+    return (<Layout>
+        <main className={skinnyWidth}>
```
loafofpiecrust, line 12: Mind running `prettier` on the files you've changed? I use the format-on-save functionality in my editor so I never have to worry about it.

loafofpiecrust (commented):
Overall looking great and thanks for looking ahead to the registration and reset password flows!

website/src/pages/login.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+        {/* TODO: */}
+        <div>
+          <label>
+            No account? <Link href="/signup">Create account</Link>
+          </label>
+        </div>
+
+        <div className={positionButton}>
+          <input className={submitButton} type="submit" value="Login" />
```
loafofpiecrust, line 104: Same here with using two words, I think "login" is the noun and "log in" is the verb.

website/src/pages/login.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <header>
+        <h1 className={centeredColumn}>Login to Your Account</h1>
+        <h4>Login to contribute to the archive by transcribing documents, recording pronunciations, providing cultural commentary, and more.</h4>
+      </header>
+
+      <form className={centeredForm} onSubmit={e => submitLogin(e, email, password)}>
+        <div >
+          <label>
+            Email *</label>
+          <input
```
loafofpiecrust, line 71: Make sure to add `name` to labels and inputs so that they are semantically linked.

website/src/pages/login.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const LoginPage = () => {
+
+  const [email, setEmail] = useState("");
+  const [password, setPassword] = useState("");
+
+  // gets the user and its setUser from the context, to update info on the current user
+  const context = useContext(UserContext);
+
+  function submitLogin(e: FormEvent<HTMLFormElement>, username: string, password: string) {
```
loafofpiecrust, line 27: I would be inclined to move this function up to the provider and have it provide a simple `logIn` function that takes username and password, which you call here. That way, all code that calls the cognito API lives in a single place.

website/src/layout.tsx
chanmioh: Change has been made!

website/src/auth.tsx
chanmioh: I will be taking a look at the docs for this and let you know if I have questions.

website/src/client/signin.tsx
chanmioh: Aaand it's gone!

website/src/pages/login.page.tsx
chanmioh: Updated!

website/src/pages/signup.page.tsx
chanmioh: Got it, updated the path.

website/src/client/layout.tsx
chanmioh: Thanks, made the change

website/src/pages/login.page.tsx
chanmioh: I thought I'd just comment these quick changes I made to make sure I understood: I added login and logout functions to the context type definition, like so: 
image

Those functions are defined in the UserProvider, and then I created hooks to access those functions: 
image

website/src/auth.tsx
chanmioh: Done (I think) I made sure useContext() doesn't get called outside of the auth.tsx file, by creating new hooks. I also moved the login and logout functions to the UserProvider component.

website/src/pages/login.page.tsx
chanmioh: Updated!

website/src/auth.tsx
loafofpiecrust: A transitive dependency is when a package is installed through another package. So if we are `dailp` and our `package.json` says "we depend on `xxx`" and `xxx` depends on `yyy`, then `yyy` is a transitive dependency of `dailp`.

website/src/pages/login.page.tsx
loafofpiecrust: Hmm I think the `login` and `logout` functions don't need the first `event` argument since they could be called from any kind of event generally.

chanmioh: Removed that event argument!

chanmioh:
Pushed a new commit with the following:

CHANGES:
- Moved Login and Logout functions from `login.page.tsx` to `auth.tsx`, and integrated them into a single useUser() hook.
- Added a popover when selecting "Log out" to avoid accidental misclicks.
image

- Replaced old HTML form with Reakit form.
- Added a barebones reset password flow, which can be access by clicking "Reset password" below the login form, or going to `/reset-password`
image

User will be prompted to enter a valid email address used to register for an account:
image

Then, theoretically they should receive a verification code in their inbox that will allow them to complete this change password form. However, I have not been able to test yet as I cannot access the test account's emails.
image

- Removed the incomplete sign up flow (for now) as it is still wip.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "amazon-cognito-identity-js": "^5.2.9"
+  }
```
loafofpiecrust, line 4: Ah sorry I wasn't specific! I meant for you to run this under `website` only, not the top-level project. We want this to get added to `website/package.json`

website/src/auth.tsx
```diff
@@ -1,53 +1,132 @@
...
+      },
+      newPasswordRequired: (data: CognitoUserSession) => {
+        console.log("New password required. Result: ", data)
+        alert("New password is required")
+      },
+    })
+  }
+
+  function logoutUser(user: CognitoUser) {
+    user?.signOut()
```
loafofpiecrust, line 102: Would it make sense to just use the `user` object already in scope instead of needing a function argument?

website/src/pages/login.css.ts
```diff
@@ -0,0 +1,40 @@
...
+  backgroundColor: "white",
+  padding: "16px",
+  borderRadius: "10%",
+  textAlign: "center",
+})
+
+export const skinnyWidth = style([
+  edgePadded,
+  centeredColumn,
+  { margin: "auto", width: "35%" },
```
loafofpiecrust, line 23: Check this out on mobile, it's super skinny. You should check out the `fullWidth` style we use elsewhere which will make an element full width on mobile but only a max pixel size on desktop. I'd start with that.

website/src/pages/login.css.ts
```diff
@@ -0,0 +1,40 @@
...
+  flex: 5,
+  padding: "8px",
+})
+
+export const logoutPopover = style({
+  display: "flex",
+  flexDirection: "column",
+  backgroundColor: "white",
+  padding: "16px",
+  borderRadius: "10%",
```
loafofpiecrust, line 16: Avoid percentages for this, check the `sprinkles.css.ts` file for some value presets we have. In there, we have `vspace` for vertical padding/margins, `hspace` for horizontal padding/margins, `radii` for `borderRadius` values, `colors` for `color` or `backgroundColor`, etc. I'd replace many of the values you have here with some of those. i.e. `borderRadius: radii.medium` perhaps. Feel free to add more values to those objects as you see fit though!

website/src/pages/login.css.ts
```diff
@@ -0,0 +1,40 @@
...
+  flexDirection: "column",
+  width: "100%",
+})
+
+export const loginFormBox = style({ width: "100%", padding: "1%" })
+
+export const positionButton = style({
+  display: "flex",
+  justifyContent: "flex-end",
+  marginTop: "5%",
```
loafofpiecrust, line 37: Very uncommon to use percentages for margins and padding. Use `vspace` or `hspace` values like I suggest above.

website/src/pages/login.page.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const loginForm = useFormState({
+    values: { username: "", password: "" },
+    onValidate: (values) => {
+      if (!values.username || !values.password) {
+        throw (
+          (!values.username && { username: "A username is required" }) || {
+            password: "A password is required",
+          }
+        )
+      }
```
loafofpiecrust, line 70: I think this could just be

```ts
if (!values.username) {
  throw ...
} else if (!values.password) {
  throw ...
}
```

Or perhaps something like this, to show both messages if both are missing:

```ts
if (!values.username || !values.password) {
  throw { username: !values.username && "...", password: !values.password && "..." }
}
``

File: website/src/pages/login.page.tsx [hunk 19]
```diff
@@ -0,0 +1,176 @@
...
+  const popover = usePopoverState()
+
+  return (
+    <>
+      <PopoverDisclosure {...popover}>Log out</PopoverDisclosure>
+      <Popover {...popover} tabIndex={0}>
+        <div className={logoutPopover}>
+          <div>Log out of DAILP?</div>
+          <Link
+            href=""
```
loafofpiecrust, line 153: Hmm we have a `Button` component and you should be able to override the styles if that's what you wanted here. I don't remember if we have a pre-built plainer button that is just text. We shouldn't use a `Link` for this.

loafofpiecrust (commented):
Adding some more comments for now. Would you be able to split the `reset-password` stuff into a separate branch so that we could move forward with the `login` page before finishing `reset-password`?

website/src/auth.tsx
chanmioh: That does make sense - let me go and change that.

website/src/pages/login.page.tsx
chanmioh: Modified!

website/src/pages/login.page.tsx
chanmioh: Done, went ahead and used a div styled with the cleanButton instead!

website/src/pages/login.css.ts
chanmioh: Got it, modified some things so on mobile it shouldn't look as wonky.

website/src/pages/login.page.tsx
loafofpiecrust: If it's semantically a button then the html element should always be `button`. Turning div into a button is also an antipattern unless there's a specific reason.

chanmioh:
> Adding some more comments for now. Would you be able to split the `reset-password` stuff into a separate branch so that we could move forward with the `login` page before finishing `reset-password`?

Definitely, and I can go ahead and upload the final code I have for this branch to review.

website/src/auth.tsx
```diff
@@ -1,53 +1,95 @@
...
-          setUser(await Auth.currentUserPoolUser())
-          break
-        case "signOut":
-          setUser(null)
-          break
-      }
+  // refreshes last authenticated user
+  useEffect(() => {
+    if (user != null) {
+      user.getSession(function (err: Error, result: CognitoUserSession | null) {
```
loafofpiecrust, line 34: I'm confused, why are you calling this function then doing nothing with the result? Just to log when you're automatically signed in?

chanmioh: While CognitoPool.getCurrentUser() gets the last authenticated user from the userpool, if the page is refreshed, the localStorage is cleared along with the user's tokens, so I added a UseEffect on component mount that will refresh the user's tokens (and restore them to the local storage). The refresh of the tokens is done automatically when the CognitoUser.getSession() method is called.

website/src/pages/login.css.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const skinnyWidth = style([
+  fullWidth,
+  edgePadded,
+
+  {
+    margin: "auto",
+    "@media": {
+      [mediaQueries.medium]: {
+        width: "35vw",
```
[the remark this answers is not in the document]
chanmioh: That makes a lot of sense, thank you!

PR #188, opened 2022-06-21
Restructure front-end style utils and add components folder

loafofpiecrust:
- Add `website/src/components` to hold general reusable components, and move several existing ones there.
- Add `website/src/style` for the theme contract and any functions or classes we use for many different styles
- Fix a bunch of tiny styling inconsistencies
- Restyle `Button` inspired by buttons on [ravelry.com](https://ravelry.com)
- Consolidate hover and focus styles to work properly for mouse, keyboard, and touch inputs. Add `onHover` and `onFocus` helper functions for writing these styles.
- Replace `a` with `Link`
- Replace re-implemented icon buttons with our standard `IconButton` component.
- Remove `sprinkles` entirely because it's less type-safe and more confusing than just making heavy use of helper functions


loafofpiecrust:
@patelShas and @chanmioh Could I get your feedback (if any) on this change before I merge it in?

chanmioh:
Looks great! I think removing sprinkles and adding `website/src/style` makes more sense - also really like having the reusable, consistent components for basics such as Link and Button. My only question would be does this majorly affect the current files in the codebase utilizing preexisting frontend imports such as from `sprinkles.css.tsx`? And will any changes need to be made to the current structure of those files?

loafofpiecrust:
> does this majorly affect the current files in the codebase utilizing preexisting frontend imports such as from `sprinkles.css.tsx`

Yes, this PR changes all imports from `sprinkles.css` to use these new files.

PR #189, opened 2022-06-23
[DP-60] Login UI - Reset Password Flow

chanmioh:
### CHANGES:
- Modified `auth.tsx`:
     - Now holds new Context properties: `authenticated` and `setAuthenticated`, which allows distinguishing between a logged in/authenticated user and a user who is in the process of resetting their password
     - New user operations
          - `resetPassword`, initiates the password reset flow by sending user an email with a verification code
          - `changePassword` verifies submitted verification code and confirms the new password
- Implemented `reset-password.page.tsx` that contains the React components for rendering a Reset Password page and a Change Password page with forms that receives user input and calls the corresponding methods in `auth.tsx`

- Reset Password page:
image

- Confirm Password page:
image

- Modified `apollo.ts` to request the /graphql-edit endpoint if user is logged in and the current deployment environment is not local
- Modified the `LoginHeaderButton` component to verify that the user is authenticated before displaying a login or logout button
- Made some small tweaks to `login.css.tsx` styles to make more concise


website/src/auth.tsx
```diff
@@ -1,31 +1,35 @@
...
   UserPoolId: process.env["DAILP_USER_POOL"] ?? "",
   ClientId: process.env["DAILP_USER_POOL_CLIENT"] ?? "",
 })
 
 export const UserProvider = (props: { children: any }) => {
+  const [authenticated, setAuthenticated] = useState(false)
+
   const [user, setUser] = useState<CognitoUser | null>(
-    userPool.getCurrentUser()
+    authenticated ? userPool.getCurrentUser() : null
```
loafofpiecrust, line 32: This gets rid of persisted logins, I'm not sure how this is the solution?

chanmioh: Thanks for the catch!! For some reason, I was worried userPool.getCurrentUser() would return a user even if they were not authenticated (i.e. started their resetting password flow) so I would check for authenticated, but realize it will always be set to false if the component is refreshed. I'm committing some new changes to fix this.

chanmioh:
### CHANGES: 

- Modified `auth.tsx`: 
     - Removed `authenticated ?` check on useState for a user
     - Added check in useEffect to make sure if there is no last authenticated user but authentication is set to true, to update authentication state to false
     - Added user as a dependency to useEffect to make sure updates are made if user is set to null or as existing

loafofpiecrust:
Looks like logins persist properly now, but I'm still wondering: why do we need both the `user` and `authenticated` states? I'm not sure when we'd ever be in a state where there's no user but you are auth'ed, or there IS a user but you _aren't_ authed. Maybe that second case is possible? Any way we can keep it in a single piece of state?

chanmioh:
### CHANGES

Overall just made some small changes to fix merge conflicts with main.

- Fixed up imports to reflect new css changes in main
- `login.css.tsx`: Increased space between Login button and form fields
- `login.page.tsx`: Created a Login Page template to keep styling consistent between pages related to authentication

chanmioh:
### CHANGES
- `auth.tsx`: Removed `authenticated` and `setAuthenticated` from `UserContextType` and now we grab the current user's credentials using the `useCredentials` hook to check if they are authenticated
- `reset-password.page.tsx`: Removed any references to the `authenticated` state to utilize useCredentials() instead, and also added some comments (hopefully they are helpful)
- `login.page.tsx`: Simplified Log out? confirmation popover, see image below:
image
- `login.css.ts`: Added styling for the newly implemented popover button

website/src/apollo.ts
```diff
@@ -6,12 +6,13 @@ import {
...
   makeOperation,
 } from "urql"
+import { Environment, deploymentEnvironment } from "./env"
 
 export const apolloClient = (token: string | null) =>
   createClient({
     url:
       process.env["DAILP_API_URL"] +
-      (token && process.env["NODE_ENV"] === "production"
+      (token && deploymentEnvironment !== Environment.Local
```
loafofpiecrust, line 15: Now that we have `graphql-edit` locally you can get rid of this extra condition

website/src/pages/login.page.tsx
```diff
@@ -10,50 +9,30 @@ import {
...
-}
-
-export const FormFields = ({
-  form,
-  name,
-  label,
-  type,
-  placeholder,
-}: FormFieldsType) => {
+export const LoginPageTemplate = (props: { header: any; content: any }) => {
```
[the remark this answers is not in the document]
chanmioh: Like this? 

image

I'm not sure if the type should be `ReactChildren` or `ReactChild`, as I get errors when I set it to `ReactChildren`:

image

website/src/pages/reset-password.page.tsx
```diff
@@ -0,0 +1,164 @@
...
+          />
+
+          <div className={positionButton}>
+            <FormSubmitButton {...form} as={Button} className={loginButton}>
+              Confirm
+            </FormSubmitButton>
+          </div>
+        </Form>
+      }
+    ></LoginPageTemplate>
```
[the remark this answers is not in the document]
chanmioh: With the addition of the `children` prop, would it be better for me to keep the closing component, so I can include the children elements in-between the opening and closing tag of `LoginPageTemplate`?

website/src/auth.tsx
```diff
@@ -68,8 +71,49 @@ export const UserProvider = (props: { children: any }) => {
...
+        alert(err.message)
+      },
+    })
+  }
+
+  function changePassword(verificationCode: string, newPassword: string) {
+    user?.confirmPassword(verificationCode, newPassword, {
+      onSuccess(data: string) {
+        setUser(null) // since user successfully changed password, reset current user's state
+        window.location.href = "/login" // redirect to login page
```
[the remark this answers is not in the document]
loafofpiecrust: Yeah so since we are in a react app with client side routing, avoiding a full page load means it doesn't have to reparse and reload the react app. Client-side navigation is much faster.

website/src/pages/login.page.tsx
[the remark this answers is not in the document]
loafofpiecrust: Yeah like that, I think. Maybe try using ReactNode type for children too. You could try ReactChild but I think that might not allow multiple child components.

website/src/pages/reset-password.page.tsx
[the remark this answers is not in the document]
loafofpiecrust: Yes!

website/src/auth.tsx
```diff
@@ -68,8 +72,51 @@ export const UserProvider = (props: { children: any }) => {
...
+      },
+    })
+  }
+
+  function changePassword(verificationCode: string, newPassword: string) {
+    user?.confirmPassword(verificationCode, newPassword, {
+      async onSuccess(data: string) {
+        setUser(null) // since user successfully changed password, reset current user's state
+        const navigationPromise = navigate("/login")
+        await navigationPromise
```
loafofpiecrust, line 100: You can reduce this to `await navigate("/login")`

website/src/urql.ts
```diff
@@ -6,14 +6,11 @@ import {
   fetchExchange,
   makeOperation,
 } from "urql"
+import { Environment, deploymentEnvironment } from "./env"
 
-export const apolloClient = (token: string | null) =>
+export const urqlClient = (token: string | null) =>
```
loafofpiecrust, line 11: Sorry I know this is annoying but since you're renaming this in this PR let's actually go for a more generic name because who knows if we'll switch again later. I'm thinking `graphqlClient` and the file name could be `website/src/graphql/client.ts` since we already have the `graphql` folder. Or if you have a better idea.

loafofpiecrust (approved):
Looks good to me, final comments!

website/src/urql.ts
chanmioh: Makes sense and I think this is a good idea! I went ahead and changed the names.

chanmioh:
### CHANGES
- `auth.tsx`: Removed a variable instantiation of `navigate` function to just call it directly
- `urql.ts`: Renamed to `client.ts` and moved to `src/website/graphql` folder, also modified anywhere that calls this file

PR #190, opened 2022-06-24
[DP-152] Document Slices

KushP10:
Added new graphql resolver,
added new parameters to form field in sql,
added new query
fixed syntax errors from new additions

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "git.ignoreLimitWarning": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "git.ignoreLimitWarning": true
-}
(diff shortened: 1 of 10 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "git.ignoreLimitWarning": true
+}
```
loafofpiecrust, line 3: You should remove this file and add `.vscode` to your global gitignore file.

types/src/document.rs
```diff
@@ -146,11 +146,16 @@ impl AnnotatedDoc {
 
     /// All the words contained in this document, dropping structural formatting
     /// like line and page breaks.
-    async fn forms(&self, context: &async_graphql::Context<'_>) -> FieldResult<Vec<AnnotatedForm>> {
+    async fn forms(
+        &self,
+        context: &async_graphql::Context<'_>,
+        start_index: Option<i64>,
+        end_index: Option<i64>,
```
loafofpiecrust, line 153: I might prefer names like `start` and `end` or `first` and `last` (although that implies inclusive on both ends). Your call though.

loafofpiecrust (approved):
Looking pretty good to me! If you've tested it and it works then I like it.

PR #192, opened 2022-06-24
Add /graphql-edit endpoint to local server with fake user

loafofpiecrust:
- Adds `/graphql-edit` to local server.
- Playground at `http://localhost:8080/graphql-edit`
- Automatically authenticated with fake user with email address `test@dailp.northeastern.edu` for now, since we don't do anything with the email.
- Fake user is in the `Editor` group, which is our only defined user group at the moment

### DIFF
--- a/graphql/src/query.rs
+++ b/graphql/src/query.rs
@@ -307,6 +307,14 @@ pub struct UserInfo {
     )]
     groups: Vec<UserGroup>,
 }
+impl UserInfo {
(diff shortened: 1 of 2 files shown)


chanmioh:
Went ahead and ran the changes locally and everything looks great! I am able to successfully fetch from the `/graphql-edit` endpoint once logged in (see image below) and access the `http://localhost:8080/graphql-edit` playground. 

image

PR #193, opened 2022-06-24
[DP-63] Add a GraphQL mutation to edit a word

chanmioh:
### CHANGES

- Defined a new InputObject to use with queries called AnnotationFormInput in `form.rs` that contains a word’s id and source_text 
- Added a new mutation in `query.rs` that takes in an AnnotationFormInput consisting of a word’s id and its new source_text field
- Added SQL call to update source text in `database_sql.rs` and added new SQL query called `update_source.sql` 
- Cleaned up `login.css.ts`, `login.page.tsx`, and `reset-password.page.tsx` styling a little bit

### DIFF
--- a/graphql/src/query.rs
+++ b/graphql/src/query.rs
@@ -5,8 +5,8 @@ use itertools::Itertools;
 use {
     dailp::async_graphql::{self, dataloader::DataLoader, Context, FieldResult, Guard},
     dailp::{
-        AnnotatedDoc, CherokeeOrthography, Database, MorphemeId, MorphemeReference, TagForm,
-        WordsInDocument,
+        AnnotatedDoc, AnnotatedFormUpdate, CherokeeOrthography, Database, MorphemeId,
+        MorphemeReference, TagForm, WordsInDocument,
     },
     serde::{Deserialize, Serialize},
     serde_with::{rust::StringWithSeparator, CommaSeparator},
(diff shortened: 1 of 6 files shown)


graphql/src/query.rs
```diff
@@ -288,6 +289,18 @@ impl Mutation {
             .await?;
         Ok(true)
     }
+
+    async fn update_source_text(
```
loafofpiecrust, line 293: Hm I think the name `update_word` is better since we'll support more fields in the future. For new things I want to use the term `word` instead of `form` because it's just more obvious to people.

graphql/src/query.rs
```diff
@@ -288,6 +289,18 @@ impl Mutation {
...
+        &self,
+        context: &Context<'_>,
+        annotation_form: AnnotationFormInput,
```
loafofpiecrust, line 296: I would probably rename this argument to `word` and the type to `AnnotatedFormUpdate` or `UpdateAnnotatedForm` (just to keep consistent with the type it's based on).

graphql/src/query.rs
chanmioh: That makes sense, I've went ahead and made the change!

types/queries/update_word.sql
```diff
@@ -0,0 +1 @@
+update word set source_text = $2 where id = $1
```
loafofpiecrust, line 1: You'll need to run `dev-generate-types` since you made a SQL change.

chanmioh:
### CHANGES

- `query.rs`: Added guard to `update_word` function so only Editor group can access this resolver. Also just modified param names from `annotation_form` to `word` to keep it consistent

loafofpiecrust (approved):

PR #199, opened 2022-07-08
[DP-106] Migrate edited collections data

hazelyn11:
Created structs to represent a Collection and a Chapter, created a function in SheetResult to parse the Collection Index directly into a Collection based on the mockup, wrote queries to insert Collections and Chapters (as a group of all chapters associated with a single Collection, Intro and Genre) into the database, added a migration step.


types/src/collection.rs
```diff
@@ -0,0 +1,15 @@
+use uuid::Uuid;
+use {
+    crate::*,
+};
+
+#[derive(Debug, Clone)]
+pub struct Collection {
+    pub title: String,
+
+    pub wordpress_menu_id: i64,
```
patelShas, line 10: Maybe make this a `Option<i64>` since like `wordpress_id` in chapter the sql allows null as a possibility.

types/src/database_sql.rs
```diff
@@ -226,6 +225,83 @@ impl Database {
...
+        .await?;
+        
+        let mut chapter_stack = Vec::new();  
+        let initial_tuple = (0, slug.clone());
+        chapter_stack.push(initial_tuple);
+
+        for current_chapter in chapters {
+
+            // Get chapter Doc ID from document title --> do in query
+            let chapter_doc_name = current_chapter.document_title;
```
patelShas, line 263: The document title does get used but the comment is a bit confusing, making it seem as if `chapter_doc_name` is supposed to be the Doc ID. The query itself gets id from the title like the comment says, but since that happens in the query it seems odd to say that here. You can just say we need the document title, not the document ID.

types/src/database_sql.rs
```diff
@@ -226,6 +225,83 @@ impl Database {
...
+            
+            // Use stack to build chapter slug
+            let mut before_chapter_index = chapter_stack.last().clone().unwrap().0;
+            while before_chapter_index != (current_chapter.index_in_parent - 1)  {
+                let last_chapter = chapter_stack.last().unwrap();
+                let last_chapter_index = last_chapter.0;
```
patelShas, line 269: You could condense this, `last_chapter` doesn't get referenced again.

migration/src/spreadsheets.rs
```diff
@@ -130,6 +132,68 @@ impl SheetResult {
...
+    /// which contains two Chapter Groups, each of which 
+    /// has a list of associated Chapters.
+    pub fn into_collection_index(self,
+        self_title: &String, 
+        self_wordpress_menu_id: &i64) -> Result<Collection> {
+        let mut chapters = Vec::new();
+        let mut self_intro_chapters = Vec::new();
+        for row in self.values.into_iter().skip(2){
+            if row[0].is_empty(){
+                self_intro_chapters = chapters.clone();
```
loafofpiecrust, line 145: You're doing a ton of cloning in this function as well, which you shouldn't need to do. You can do something like this instead:

```rs
let row = row.into_iter();
let first_value = row.next().ok_or_else(|| anyhow!("Missing first value"))?;
let second_value = row.next().ok_or_else(|| anyhow!("Missing second value"))?;
```
It's not the shortest but it's more idiomatic.

migration/src/spreadsheets.rs
```diff
@@ -130,6 +132,68 @@ impl SheetResult {
...
+                let mut doc_string= None;
+
+                let wp_id_if_applicable = row[5].clone();
+                let mut wp_id = None;
+
+                if wp_id_if_applicable.is_empty(){
+                    wp_id = None;
+                }
+                else {
+                    wp_id = Some(wp_id_if_applicable.parse::<i64>().unwrap());
```
loafofpiecrust, line 162: Avoid using `.unwrap()` whenever possible because if you unwrap a `None` it will crash the whole program (panic) instead of returning an `Error` from this function. You can do `Some(x.parse()?)` (if you want a failed parse to return an error) or `x.parse().ok()` (if you want a failed parse to result in no value for `wp_id`).

migration/src/spreadsheets.rs
```diff
@@ -130,6 +132,68 @@ impl SheetResult {
...
+                    index_in_parent: index_i64,
+                    url_slug: row[1].clone(),
+                    chapter_name: row[2].clone(),
+                    document_title: doc_string,
+                    id: None,
+                    document_id: None,
+                    wordpress_id: wp_id,
+                };
+
+                chapters.push(new_chapter.clone());
```
loafofpiecrust, line 182: Why did you think you needed to call `clone()` here in the first place? `new_chapter` is created immediately before and not used anywhere else. Maybe what clone does is unclear?

types/src/chapter.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub index_in_parent: i64,
+
+    pub chapter_name: String,
+
+    pub document_title: Option<String>,
+
+    pub document_id: Option<DocumentId>,
+
+    pub wordpress_id: Option<i64>,
+}
```
loafofpiecrust, line 21: Make sure to run the code formatter. Outside of your editor, you can run `cargo fmt` to do this across the project.

migration/src/edited_collection.rs
```diff
@@ -0,0 +1,28 @@
...
+use dailp::chapter::Chapter;
+use dailp::collection::Collection;
+
+pub async fn migrate_edited_collection(db: &Database) -> anyhow::Result<()> {
+    let res = SheetResult::from_sheet("12R07Ks8A5g2jffqoILJw7nRn6GehwNwdwosxaJBwNNM", None)
+        .await?
+        .into_collection_index(&"Test".to_string(), &4579)?;
+    let collection = res;
+
+    let collection_slug = "test_sheet".to_string();
```
loafofpiecrust, line 13: This can be `cwkw`

migration/src/edited_collection.rs
```diff
@@ -0,0 +1,28 @@
+use crate::spreadsheets::SheetResult;
+use dailp::{Database};
+use anyhow::Result;
+use dailp::chapter::Chapter;
+use dailp::collection::Collection;
+
+pub async fn migrate_edited_collection(db: &Database) -> anyhow::Result<()> {
+    let res = SheetResult::from_sheet("12R07Ks8A5g2jffqoILJw7nRn6GehwNwdwosxaJBwNNM", None)
+        .await?
+        .into_collection_index(&"Test".to_string(), &4579)?;
```
loafofpiecrust, line 10: This can be `Cherokees Writing the Keetoowah Way`

migration/src/spreadsheets.rs
```diff
@@ -130,6 +132,68 @@ impl SheetResult {
...
+                let index_in_parent_string = row[0].clone();
+                let index_i64 = index_in_parent_string.parse::<i64>().unwrap();
+
+                let document_if_applicable = row[3].clone();
+                let mut doc_string= None;
+
+                let wp_id_if_applicable = row[5].clone();
+                let mut wp_id = None;
+
+                if wp_id_if_applicable.is_empty(){
```
loafofpiecrust, line 158: This is an anti-pattern in rust. if-elses are expressions, so you can write:
```rs
let wp_id = if ... { x } else { y };
```
This also applies to `doc_string`

types/queries/insert_all_chapters.sql
```diff
@@ -0,0 +1,14 @@
...
+    title, 
+    document_id, 
+    wordpress_id, 
+    index_in_parent,
+    chapter_path)
+values (
+    $1, 
+    (select id
+    from document
+    where title = $2),
```
loafofpiecrust, line 11: You can simplify this pattern by inserting the result of a select directly:
```sql
insert into ... (...)
select $1, document.id, $3, $4, $5
from document
where title = $2
```

by the way, you don't want to use `document.title`, you want to use `document.short_name` because that's `WJ36` instead of `Story of the Switch Striker` for example.

types/queries/insert_all_chapters.sql
```diff
@@ -0,0 +1,14 @@
...
+    index_in_parent,
+    chapter_path)
+values (
+    $1, 
+    (select id
+    from document
+    where title = $2), 
+    $3, 
+    $4,
+    text2ltree ($5))
```
loafofpiecrust, line 14: Did you try without `text2ltree` and find that you do actually need it? I didn't think you would need it.

types/src/chapter.rs
```diff
@@ -0,0 +1,21 @@
+use uuid::Uuid;
+use {
+    crate::*,
+};
+
+#[derive(Debug, Clone)]
+pub struct Chapter {
```
loafofpiecrust, line 7: Can we either call this `CollectionChapter` or have it live in the `collection` module and use it like `collection::Chapter` (that might be better)? They're very closely associated data and I want to make that clear in the naming.

migration/src/edited_collection.rs
```diff
@@ -0,0 +1,28 @@
...
+    let collection = res;
+
+    let collection_slug = "test_sheet".to_string();
+
+    // Insert collection into database-- hardcode CWKW stuff/move outside of this function?
+    db.upsert_collection(collection.title,collection.wordpress_menu_id, collection_slug.clone()).await?;
```
loafofpiecrust, line 16: You wrote and instantiated the `Collection` data type, why not pass that in here as the sole argument? Also, I think the `slug` should be a field on `Collection`, that matches the DB type as well.

types/src/database_sql.rs
```diff
@@ -226,6 +225,83 @@ impl Database {
...
+        &self,
+        chapters: Vec<Chapter>,
+        slug: String) -> Result<String>{
+
+        let mut tx = self.client.begin().await?;
+
+        // Delete all chapter data before re-inserting
+        query_file!(
+            "queries/delete_chapters.sql")
+        .execute(&mut tx)
```
loafofpiecrust, line 253: Jesus, no, don't delete _all_ chapters. This function needs to be scoped to a specific collection, so that you only delete the chapters under `cwkw` then insert all the `cwkw` chapters.

types/src/database_sql.rs
```diff
@@ -226,6 +225,83 @@ impl Database {
...
+            // Concatenate URL Slugs of all chapters on the path
+            let mut chapter_stack_cur = chapter_stack.clone();
+            let mut url_slug_cur = current_chapter.url_slug.clone();
+            let final_path = (current_chapter.index_in_parent, current_chapter.url_slug.clone());
+            chapter_stack.push(final_path);
+
+            while !(chapter_stack_cur.is_empty()){
+                let temp_chapter = chapter_stack_cur.pop().unwrap();
+                let cur_slug = temp_chapter.1;
+                url_slug_cur = cur_slug + "." + &url_slug_cur;
```
loafofpiecrust, line 285: This is more idiomatic and efficient in Rust:
```rs
format!("{}.{}", cur_slug, url_slug_cur)
```

types/src/database_sql.rs
```diff
@@ -226,6 +225,83 @@ impl Database {
...
+            }
+
+             // Insert chapter data into database
+             query_file!(
+                 "queries/insert_all_chapters.sql",
```
loafofpiecrust, line 290: This is called `insert_all_chapters` but it appears to only insert one chapter. There are a bunch of queries that insert multiple rows, it's a bit weird to read but to find them just search for `unnest`. On the rust end, that means passing one argument per field where each is a list. So instead of this:
```
insert(a1, b1, c1);
insert(a2, b2, c2);
```

You are doing this:
```
insert_all([a1, a2], [b1, b2], [c1, c2])
```

types/src/chapter.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub id: Option<Uuid>,
+    
+    pub url_slug: String,
+
+    pub index_in_parent: i64,
+
+    pub chapter_name: String,
+
+    pub document_title: Option<String>,
```
loafofpiecrust, line 16: We don't need this field at all.

migration/src/spreadsheets.rs
```diff
@@ -130,6 +132,68 @@ impl SheetResult {
...
+                    doc_string = Some(document_if_applicable);
+                }
+
+                let new_chapter = Chapter {
+                    index_in_parent: index_i64,
+                    url_slug: row[1].clone(),
+                    chapter_name: row[2].clone(),
+                    document_title: doc_string,
+                    id: None,
+                    document_id: None,
```
loafofpiecrust, line 178: ??? You'll need the document's `short_name` field here, which perhaps I should have called slug. But anyway, so you could call it `document_short_name: String` perhaps since you don't have the document's `id` yet (`DocumentId` is a wrapper for UUID) and we don't need the `title` field.

migration/src/edited_collection.rs
```diff
@@ -0,0 +1,28 @@
...
+    // Append both intro and genre chapters together
+    // to add all chapters to the database at the same time
+   
+    let mut chapter_group_one = collection.intro_chapters.clone();
+    let mut chapter_group_two = collection.genre_chapters.clone();
+
+    chapter_group_one.append(&mut chapter_group_two);
+
+    db.insert_all_chapters(chapter_group_one, collection_slug.clone()).await?;
```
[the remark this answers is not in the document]
hazelyn11: Fixed, thanks!

migration/src/spreadsheets.rs
hazelyn11: Cool, fixed this, and used this logic to remove the if/else statements, thanks!

migration/src/spreadsheets.rs
hazelyn11: No, yeah, I understand why clone() is not necessary here-- I think it may have been leftover from before, when I was trying to build the chapter path in this function as well and doing unnecessarily complicated and convoluted things with the chapter data as a result.

migration/src/spreadsheets.rs
hazelyn11: Thanks! I realized I didn't even need the if-else statements, so I just rewrote the code without them, but I will keep this in mind for the future.

types/queries/insert_all_chapters.sql
hazelyn11: yes -- when I remove text2ltree I get this error: expected struct `PgLTree`, found struct `std::string::String`

types/queries/insert_all_chapters.sql
hazelyn11: Hmm, I tried rewriting it the way you suggested, but that meant that only chapters with an associated document ended up being inserted. And I can change it so that we're using the document short name-- the reason I had it the other way was since in the draft spreadsheet, the short name was marked as optional to include, and it was the long title that would be required if there was a document associated.

Maybe I misunderstood, in which case I can change it so it's reflected as short name required/longer title optional in the spreadsheet as well?

types/src/chapter.rs
hazelyn11: Yes, that makes sense! The Chapter type now lives in the Collection module.

migration/src/edited_collection.rs
hazelyn11: Right, this makes sense! And I added slug as a field-- not sure why I hadn't done that before

types/src/database_sql.rs
hazelyn11: Right, I kind forgot that we would eventually have multiple edited collections so that's why I had done that-- what would the best way to go about doing this be? I was thinking since each chapter path would have the collection slug included in it, I could just delete every chapter where the collection slug forms part of the chapter path, but I'm not sure if there's an easier/simpler/better way to do that.

migration/src/spreadsheets.rs
hazelyn11: Oh, okay-- I'll change the Chapter struct to reflect this, and make the Document Title field optional on the spreadsheet, and the Document ID field required (if applicable) on the spreadsheet to reflect this as well.

hazelyn11: Fixed so now the short name is used to get the document ID instead of the document title.

types/src/database_sql.rs
hazelyn11: Could I just rename the query I wrote so it's not misleading, or should I make a new query which uses unnest?

loafofpiecrust: I haven't checked yet if you addressed this already, but for now you could rename it since we aren't inserting all that many chapters yet.

types/src/database_sql.rs
loafofpiecrust: Yeah that's exactly what I'd do, delete chapters where the very first element in the path is exactly the collection slug. The [ltree docs](https://www.postgresql.org/docs/current/ltree.html) show some examples of these queries. I _think_ the best way to do this exact query (rather than just checking if `cwkw` is any of the ancestors or just somewhere in the path), is: `chapter.path ~ format('%s.*', $1)`

That checks that `chapter.path` matches the query "any path that starts with the string $1 and has anything after that".

types/queries/insert_all_chapters.sql
loafofpiecrust: Oh! That means sqlx has a Rust type for ltrees, which I didn't know. Just quickly searched [their docs and found it](https://docs.rs/sqlx/latest/sqlx/postgres/types/struct.PgLTree.html). I would suggest using that type because then we confirm that our string is a valid path before even getting to the DB. Check out [this `from_str` function](https://docs.rs/sqlx/latest/sqlx/postgres/types/struct.PgLTree.html#impl-FromStr) which you should be able to use like this:

```rs
PgLTree::from_str(s)?
```

types/queries/insert_all_chapters.sql
loafofpiecrust: Aha you are right about the optionality of the linked document, so yeah what you wrote seems best!

I repeat, we should remove the full document title from the spreadsheet entirely. The chapter itself has a title already and we don't need to repeat the document title on this table.

migration/src/spreadsheets.rs
loafofpiecrust: Gotcha all good!

types/src/database_sql.rs
loafofpiecrust: I bet you could come up with another way of writing that query that doesn't use `format` too.

migration/src/main.rs
```diff
@@ -26,25 +26,25 @@ async fn main() -> Result<()> {
 
     let db = Database::connect().await?;
 
-    println!("Migrating Image Sources...");
-    migrate_image_sources(&db).await?;
+    // println!("Migrating Image Sources...");
+    // migrate_image_sources(&db).await?;
```
loafofpiecrust, line 30: Make sure to drop these comments

migration/src/spreadsheets.rs
```diff
@@ -137,49 +137,37 @@ impl SheetResult {
...
+                let wp_id = (&cur_row[4]).parse::<i64>().ok();
+                let doc_string = Some(cur_row[5].to_string());
 
                 let new_chapter = Chapter {
                     index_in_parent: index_i64,
-                    url_slug: row[1].clone(),
-                    chapter_name: row[2].clone(),
-                    document_title: doc_string,
+                    url_slug: (&cur_row[1]).to_string(),
+                    chapter_name: (&cur_row[2]).to_string(),
```
loafofpiecrust, line 168: Hmm doing `(&x).to_string()` is the same as doing `x.clone()` but with more steps, they both allocate. You can't move values out of a `Vec` using the indexing syntax `list[i]`, that's always read-only.  You could do exactly what you did earlier in this function:

```rs
let row_values = cur_row.into_iter();
// you could also use .ok_or_else(...)
let index: i64 = row_values.next().unwrap().parse()?;
let url_slug = row_values.next().unwrap();
// and so forth
```

This and your current code will panic if a row doesn't have enough values. You should check at the beginning `cur_row.len() > 4` or however many values it needs, then using `.unwrap()` or `cur_row[i]` here is pretty safe.

types/src/collection.rs
```diff
@@ -7,9 +7,28 @@ use {
...
+    
+    pub url_slug: String,
+
+    pub index_in_parent: i64,
+
+    pub chapter_name: String,
+
+    pub document_short_name: Option<String>,
+
+    pub document_id: Option<DocumentId>,
```
loafofpiecrust, line 29: Why are these both fields here? I think you don't fill in `document_id` with `Some` in your code ever right?

types/queries/insert_all_chapters.sql
hazelyn11: Okay cool, I have removed the full document title from the spreadsheet.

types/queries/insert_all_chapters.sql
hazelyn11: Cool, fixed!

types/src/database_sql.rs
hazelyn11: Hmm, when I tried the expression you suggested I got errors since it interpreted format('%s.*', $1) as text-- maybe I missed it but I couldn't find a way to convert from text to lquery within a query in the documentation.

What I ended up doing was using PgLQuery::from_str outside of the query itself, after concatenating the slug with ".*". This way also doesn't use format, but I'm not sure if there's a more intuitive/better way to write it.

migration/src/spreadsheets.rs
hazelyn11: okay-- for some reason I couldn't get it to work using a check for cur_row.len() (it compiled and everything, the right values just weren't showing up in the tables). Instead, I used a peekable iterator before using .unwrap(). Would that be an okay solution?

types/src/collection.rs
hazelyn11: you're right-- document_id is not needed, got rid of it!

migration/src/spreadsheets.rs
loafofpiecrust: Hmm that might work fine, does it let you peek multiple cells or something?

hazelyn11: The last three fields of the spreadsheet are optional, so I just used peek three times before assigning data to any of those fields. (I also committed files with this change if you wanted to take a look)

types/src/database_sql.rs
loafofpiecrust: Ohhh that should work!

migration/src/spreadsheets.rs
```diff
@@ -155,20 +155,35 @@ impl SheetResult {
...
+
+                // Both of these fields are optional, and will panic if out of bounds
+                let mut wp_id = None;
+                if row_values.peek().is_some() {
+                    wp_id = row_values.next().unwrap().parse::<i64>().ok();
+                }
+
+                let mut doc_string = None;
+                if row_values.peek().is_some() {
+                    doc_string = Some(row_values.next().unwrap());
```
loafofpiecrust, line 178: Btw you can do `doc_string = row_values.next()`

types/src/database_sql.rs
```diff
@@ -244,8 +247,9 @@ impl Database {
     ) -> Result<String> {
         let mut tx = self.client.begin().await?;
 
-        // Delete all chapter data before re-inserting
-        query_file!("queries/delete_chapters.sql")
+        // Delete previous chapter data stored for a particular collection before re-inserting
+        let collection_slug = PgLQuery::from_str(&[slug.clone(), ".*".to_string()].concat())?;
```
loafofpiecrust, line 251: This should be easier to read: `format!("{}.*", slug)`

loafofpiecrust (approved):

PR #200, opened 2022-07-11
Added an editable word panel on documents

chanmioh:
### CHANGES:
- "Edit" button displays when a user is logged in
image

- Upon pressing "Edit", authenticated users view an editable form display of the word in the panel
image

- Currently, only the `Syllabary Characters` field edits can be saved

- `form.rs`: 
    - Added `id` function for an `AnnotatedForm` object to retrieve a word’s id if needed
- `queries.graphql`: 
    - Added `id` field to FormFieldsFragment to get id field for mutation queries
    - Added mutation query `UpdateWord` to call `updateWord` resolver
- `document-page.tsx`: 
    - Added `FormProvider` that contains a context for tracking the state of the current form
- `segment.tsx`: 
    - On clicking a word in the document, the current form will update to access information about that word allowing its features to be edited
    - Rerenders a word in the document and its panel information when “Save” button is chosen
- `form-context.tsx`: 
    - Creates a form state that contains a word of `FormFieldFragment` type, keeping track of the current values for a word through an editable form
    - Calls GraphQL mutation to update a word 

- `panel.tsx`: Contains components to display a normal word panel, or an editable word panel
- `panel.css.tsx`: Renamed from `word-panel.css.tsx` to keep file names consistent
- `edit-word-panel.tsx`: Displays FormInput components to modify words, and buttons to edit, cancel, and save the current form
- `edit-word-panel.css.tsx`: Contains styling for the editable word panel
- `word-panel.tsx`: Contains a <div> component that displays info about a given word’s feature


website/src/panel.tsx
```diff
@@ -0,0 +1,261 @@
...
+import * as css from "./panel.css"
+import { WordPanel } from "./word-panel"
+
+export interface PanelDetails {
+  currContents: FormFieldsFragment | null
+  setCurrContents: (currContents: FormFieldsFragment | null) => void
+}
+
+/** Displays the right-side panel information of the currently selected word. */
+export const Panel = (props: {
```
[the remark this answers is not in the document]
chanmioh: Thanks, I went ahead and changed it to `PanelLayout` if that makes more sense!

chanmioh:
> The website preview gets stuck here: image where the document segments are still loading even after other feature's like the audio recording are loaded. So, I haven't actually seen the front end functionality.
> 
> Also, just make sure that the website and new features don't glitch out when the screen isn't very wide or on mobile.

Thanks for the heads up. I'm not too sure why the website preview isn't working since it's working on my end locally. I did make some changes to the `FormFields` fragment in `queries.graphql` that might be causing issues? I'll take a look into this!

chanmioh:
> The website preview gets stuck here: image where the document segments are still loading even after other feature's like the audio recording are loaded. So, I haven't actually seen the front end functionality.
> 
> Also, just make sure that the website and new features don't glitch out when the screen isn't very wide or on mobile.

Took another look at this, and this is the error message that appears from the graphql response to get the document contents: 

image

So since I added `id` to the `FormFragments` on an `AnnotatedForm`, and modified the backend so that `AnnotatedForm` has a function for returning the `id`, I think that's why netlify is having issues displaying the document contents. Not too sure how to fix this, though, other than the modifications I made.

chanmioh:
### CHANGES
- `form.rs`, `database_sql.rs`, `update_word.sql`: Added `commentary` field to SQL update query and modified `AnnotatedFormUpdate` accordingly
- `form-context.tsx`: Added `commentary` field to mutation call
- `segment.tsx`: 
    - Added a `useState` for `isSelected` and moved checking it into a `useEffect`, as previously was giving an error for updating the currently selected word state in a non-useeffect condition
    - Added a `useEffect` for checking if the current word selected has changed
    - Added a `useEffect` for checking if the form has been submitted, and to render UI changes accordingly for a responsive UI
        - If the form has not been submitted, then return the form back to the prior segment state
- `document.page.tsx`: Added a `useEffect` for rerunning queries on edit changes so the contents contain the latest database contents
- `auth.tsx`: Modified `useEffect`’s dependency array to contain `token` rather than `user` since I realized when `token` expires, this `useEffect` should be called to refresh the token

loafofpiecrust:
@patelShas I've noted this before but when a PR includes back-end changes you cannot rely on the preview deployment quite yet. It only deploys the front-end and points to the dev endpoint at `dev.dailp.northeastern.edu/api/graphql`. So you'll need to review the changes locally.

types/src/form.rs
```diff
@@ -173,6 +178,8 @@ pub fn is_root_morpheme(s: &str) -> bool {
 pub struct AnnotatedFormUpdate {
     /// Unique identifier of the form
     pub id: Uuid,
-    /// Original source text that can be either undefined or null
+    /// Original source text that could also be undefined
```
loafofpiecrust, line 181: the "could also be undefined" is not really needed. I would just add to the docstring for the struct that all fields are optional.

website/src/auth.tsx
```diff
@@ -40,7 +42,7 @@ export const UserProvider = (props: { children: any }) => {
         }
       })
     }
-  }, [user])
+  }, [token])
```
loafofpiecrust, line 45: Why did this change exactly? Since the hook uses `user`, it becomes less easy to understand with this mismatch. If it's necessary, then add a comment in the code explaining why.

website/src/edit-word-panel.css.tsx
```diff
@@ -0,0 +1,52 @@
+import { style } from "@vanilla-extract/css"
+import { button, cleanButton } from "src/components/button.css"
+import { fonts, hspace, radii, vspace } from "src/style/constants"
+import { marginX, paddingX, paddingY } from "src/style/utils"
+
+export const formStyle = style({
```
loafofpiecrust, line 6: Stylistically, I prefer not having a `Style` suffix and instead importing this file as `css` or even `style(s)` if you like that better. Then you could do `style.form` or `css.button` instead.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,127 @@
...
+export const EditButton = () => {
+  const { form, isEditing, setIsEditing } = useForm()
+
+  return (
+    <Form {...form} className={formStyle}>
+      {isEditing ? (
+        // Displays a "Cancel" button and "Save" button in editing mode.
+        <>
+          <Button
+            className={cleanButtonStyle}
```
loafofpiecrust, line 31: I would suggest using the component from our component library instead, looks like all you changed is the font family. If you must do that, you can do `<CleanButton className={xxx}` and your `xxx` class will _add_ to, not replace the base class for that component.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,127 @@
...
+            onClick={() => {
+              setIsEditing(false)
+            }}
+          >
+            Cancel
+          </Button>
+
+          <FormSubmitButton
+            {...form}
+            className={buttonStyle}
```
loafofpiecrust, line 42: For reference if you can replace this with a preset button component, it should let you do `<XButton as={FormSubmitButton}`. It seems like our library is missing a component for a button with icon _and_ text, perhaps. If so then extract out your styles into a reusable component.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,127 @@
...
+    </Form>
+  )
+}
+
+/** Displays a FormInput with its corresponding feature data from the Reakit form. */
+const EditWordPanel = (props: { features: string[] }) => {
+  const { form } = useForm()
+
+  if (!form || !form.values.word) {
+    return <></>
```
loafofpiecrust, line 70: You can use `return null`

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,127 @@
...
+    case "simplePhonetics":
+      label = "Simple Phonetics"
+      break
+    case "romanizedSource":
+      label = "Romanized Source"
+      break
+    case "englishGloss":
+      label = "English Translation"
+      break
+    default:
```
loafofpiecrust, line 105: You've got high cohesion between your behavior and your copy/data which means a higher maintenance burden. I shouldn't need to edit the general field editor component to change the label of a specific field. I would make `label` a prop for sure.

website/src/panel-layout.tsx
```diff
@@ -0,0 +1,267 @@
...
+                </td>
+                <td className={css.editMatchingCells}>
+                  {segment.matchingTag?.__typename === "TagForm" ? (
+                    <p.panelType
+                      features={[
+                        "segments",
+                        index.toString(),
+                        "matchingTag",
+                        "title",
+                      ]}
```
loafofpiecrust, line 197: This is a really interesting solution, can you give me a rundown of the thinking behind it? I understand you're looking for a generic component to edit a field at any level of nesting, but I'm curious to hear more. How does this compare to passing a setter like `(word, value) => word.segments[index].matchingTag.title = value` or whatever?

website/src/panel-layout.tsx
```diff
@@ -0,0 +1,267 @@
...
+                    features={["segments", index.toString(), "morpheme"]}
+                  />
+                </td>
+                <td className={css.editMatchingCells}>
+                  {segment.matchingTag?.__typename === "TagForm" ? (
+                    <p.panelType
+                      features={[
+                        "segments",
+                        index.toString(),
+                        "matchingTag",
```
loafofpiecrust, line 195: Also, I think you're misunderstanding what `matchingTag` is. That's not inherent data, that's calculated as the glossary entry which provides a definition for this morpheme. Take a look at the `matching_tag` resolver in Rust, and the doc comment for it, passing this field will do nothing. I mentioned editing the morphemic segmentation as being another in-depth project because assigning morphemes a `matchingTag` means some kind of auto-complete drop-down that shows all available matching tags. That's not in scope for this PR I'd say but I would at least avoid trying to change the matchingTag fields.

website/src/word-panel.tsx
```diff
@@ -1,195 +1,23 @@
...
-            <td className={css.tableCells}>
-              {segment.matchingTag ? segment.matchingTag.title : segment.gloss}
-            </td>
-          </tr>
-        ))}
-      </table>
-    )
+  // If the feature is an english gloss, it will return a special case of the component with quotes around the text
+  if (feature === "englishGloss") {
+    return <div>‘{props.word[feature]}’</div>
```
loafofpiecrust, line 19: This is another high cohesion problem, we should absolutely not be checking raw string equality. [This article is about Swift](https://academy.realm.io/posts/tryswift-andyy-hope-swift-eye-stringly-typed-api/) but I think a good enumeration of the problems with a "stringly typed API" like this one.

website/src/auth.tsx
chanmioh: I noticed while testing out the editable word functionality, that after some period of time while logged in, the user's token will become invalid. I made this `useEffect` originally as a solution to that problem, but realized the only state that changes is `token` and not the `user` object so the dependency hook needs `token` to work properly.

types/src/form.rs
chanmioh: Gotcha, made that change!

website/src/edit-word-panel.tsx
chanmioh: Got it thanks, went ahead and added that as a prop!

website/src/word-panel.tsx
chanmioh: Makes sense, I went ahead and changed this to add the quotation marks in the `PanelContent` component instead to reduce coupling.

website/src/edit-word-panel.tsx
chanmioh: Gotcha, hope it's okay but I went ahead and changed it to an `IconButton` so there will be a gray background when hovering over the button.

website/src/panel-layout.tsx
chanmioh: I'm not sure if I understood the question 100%, but this wasn't my preferred solution, just one that worked and made the most sense to me. Since Reakit forms display a FormInput of a value with the value's key as a string, I would pass in the value's key as a list of string to identify the desired FormInput. I'm open to other solutions though as this pattern likely won't work now that I'm understanding matching tags correctly!

chanmioh:
### CHANGES

`edit-word-panel.css`: Removed style suffix from styles
`edit-word-panel`: 
	- Utilized reusable components
	- Removed switch case and added `label` prop
	- Added `input` prop as some fields are better displayed as a `textarea` (esp. commentary where there are multiple fields)
`panel-layout`: 
	- Quotation marks are appended to translation depending on the panel display mode
	- Reverted VerticalMorphemicSegmentation to previous version for now
`word-panel`: Removed if condition
`button`: Added a reusable component `IconTextButton`, a button with an icon and text

website/src/edit-word-panel.tsx
chanmioh: Cool, I went ahead and extracted it into a reusable component!

types/src/database_sql.rs
```diff
@@ -306,9 +306,14 @@ impl Database {
...
 
     pub async fn update_word(&self, word: &AnnotatedFormUpdate) -> Result<Uuid> {
-        query_file!("queries/update_word.sql", word.id, word.source.value())
-            .execute(&self.client)
-            .await?;
+        query_file!(
+            "queries/update_word.sql",
+            word.id,
+            word.source.value(),
+            word.commentary.value()
```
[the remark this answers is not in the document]
chanmioh: I think that solution makes sense. If I have any more ideas I'll let you know.

website/src/panel-layout.tsx
chanmioh: Got it, for now I'll leave that part for another time.

chanmioh:
### small CHANGES
- `segment`: Moved `useEffect` updating the right-hand word panel outside of the `AnnotatedForm` component to lift state of the word panel’s content

chanmioh:
### CHANGES
- `query.rs`: `update_word` function now takes in an AnnotatedFormUpdate object directly so certain methods can be called on the object (such as take())
- `update_word.sql`:  Updated to contain cases for an empty array, an array with a null value, and an array with a string value to update a field
    - If empty array: Field is not updated
    - If null value in an array: Field is set to null (or not updated if the field is not-nullable, such as `source_text`)
    - If defined value in an array: Field is set to that value
- `database_sql.rs`: `update_word` instantiates an array for source and commentary fields. Pushes in either a null or defined value if the given `AnnotatedFormUpdate` object is not undefined.

types/queries/update_word.sql
```diff
@@ -1 +1,13 @@
...
+update word set
+    source_text =
+        case
+            when $2::text[] != '{}' and $2[1] is not null then $2[1]
+            else source_text
+        end,
+    commentary =
+        case
+            when $3::text[][1] is not null then $3[1]
+            when $3[1] is null then null
```
loafofpiecrust, line 10: I think this can be simplified down to:

```sql
when $3::text[] != '{}' then $3[1]
```

since we just want whatever the value is if the list isn't empty.

types/src/database_sql.rs
```diff
@@ -305,10 +305,26 @@ impl Database {
...
+        let mut source: Vec<Option<String>> = Vec::new();
+        let mut commentary: Vec<Option<String>> = Vec::new();
+
+        if !word.source.is_undefined() {
+            source.push(word.source.take());
+        }
+
+        if !word.commentary.is_undefined() {
+            commentary.push(word.commentary.take());
+        }
```
loafofpiecrust, line 318: We could remove mutability here by doing it this way instead:

```rs
let source = if word.source.is_undefined() {
  Vec::new()
} else {
  vec![word.source.take()]
};
```

loafofpiecrust: Since we're gonna use this pattern a bunch, you could also abstract this into a function that goes from `MaybeUndefined<T>` -> `Vec<Option<T>>`

types/src/form.rs
```diff
@@ -130,6 +130,11 @@ impl AnnotatedForm {
     async fn document_id(&self) -> DocumentId {
         self.position.document_id
     }
+
+    /// Unique identifier of this form
+    async fn id(&self) -> Uuid {
+        self.id.unwrap_or_default()
```
loafofpiecrust, line 136: Definitely don't use `unwrap_or_default()` here because that'll return a zeroe'd out UUID that doesn't mean anything if `self.id` is `None`. Words coming from the DB should _always_ have `Some(id)`, only ones being inserted have `None`. So you can throw an error here by making the return value `FieldResult<Uuid>` and using `anyhow` to produce an error if it's `None`.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+            Cancel
+          </IconButton>
+
+          <IconTextButton
+            {...form}
+            icon={<IoCheckmarkSharp />}
+            text={"Save"}
+            className={css.editPanelButton}
+            as={FormSubmitButton}
+            aria-label="Save the selected word and its features"
```
loafofpiecrust, line 42: Why are the `aria-label`s for these buttons so long? I'm uncertain whether they need to be, since the existing text ("Save", "Edit", "Cancel") will be the default label. The extra clarification could be valuable though.

website/src/components/button.tsx
```diff
@@ -17,3 +17,25 @@ export const IconButton = ({
...
+
+type IconTextButtonProps<C extends React.ElementType> = ButtonProps & {
+  icon: ReactNode
+  text: string
+  as?: C
+}
+
+export const IconTextButton = <C extends React.ElementType>({
+  icon,
+  text,
```
loafofpiecrust, line 29: I think it'll be better to use the special `children` prop for the text.

website/src/form-context.tsx
```diff
@@ -0,0 +1,67 @@
...
+/** Instantiates a form state used to keep track of the current word and information about all its features. */
+export const FormProvider = (props: { children: any }) => {
+  const [isEditing, setIsEditing] = useState(false)
+  const word: FormFieldsFragment | undefined = undefined
+
+  const [updateWordResult, updateWord] = useUpdateWordMutation()
+
+  /** Calls the backend GraphQL mutation to update a word. */
+  const runUpdate = (variables: { word: AnnotatedFormUpdate }) => {
+    updateWord(variables).then((result) => {
```
loafofpiecrust, line 29: You can use `async` and `await` here instead of `.then`, but also please remove the extra log.

website/src/form-context.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const form = useFormState({
+    values: {
+      word,
+    },
+    onValidate: (values) => {},
+    onSubmit: (values) => {
+      if (!values.word) {
+        throw { values: { word: "No word found" } }
+      }
```
loafofpiecrust, line 42: You can also check for `values.word` existing and being valid in the `onValidate` function.

website/src/pages/documents/document.page.tsx
```diff
@@ -59,10 +60,12 @@ const AnnotatedDocumentPage = (props: { id: string }) => {
   return (
     <Layout>
       <Helmet title={doc?.title} />
-      <main className={css.annotatedDocument}>
-        <DocumentTitleHeader doc={doc} showDetails={true} />
-        <TabSet doc={doc} />
-      </main>
+      <FormProvider>
```
loafofpiecrust, line 63: This provider is really high up in the component hierarchy, I think it should live within the word details panel, since it isn't needed until that particular component.

website/src/panel-layout.tsx
```diff
@@ -0,0 +1,264 @@
...
+            onClick={() => props.setContent(null)}
+            aria-label="Dismiss selected word information"
+          >
+            <MdClose size={32} />
+          </IconButton>
+
+          {token && <EditButton />}
+        </div>
+
+        <h2 className={css.cherHeader}>{props.segment.source}</h2>
```
loafofpiecrust, line 51: The styling for this element is off, it shouldn't have the margin on the sides and it's got a big margin above it.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <Form {...form} className={css.form}>
+      {isEditing ? (
+        // Displays a "Cancel" button and "Save" button in editing mode.
+        <>
+          <IconButton
+            className={css.cancelButton}
+            round={false}
+            aria-label="Cancel editing the selected word and its features"
+            onClick={() => {
+              setIsEditing(false)
```
loafofpiecrust, line 30: I think we should hide the close button while editing, to prevent accidental loss of edits. If you disagree though, you should effectively "Cancel" if I click the close button instead of cancel or save.

website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+import * as css from "./edit-word-panel.css"
+import { useForm } from "./form-context"
+import { FormFieldsFragment } from "./graphql/dailp"
+
+/** Button that allows user to enter edit mode in the word panel, and edit fields of a word. */
+export const EditButton = () => {
+  const { form, isEditing, setIsEditing } = useForm()
+
+  return (
+    <Form {...form} className={css.form}>
```
loafofpiecrust, line 21: I think I want the header of the word details panel to stay the same as it is when you aren't logged in (Selected Word header, syllabary underneath, and close button in top-right corner) while we iterate on that design element. Then when you are logged in, this UI is fine (close button top-left, cancel and save on top-right).

website/src/panel-layout.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  // Contains components rendering a word's segments and its english translation.
+  const wordPartsContent = (
+    <>
+      <VerticalMorphemicSegmentation
+        panelType={props.panelType}
+        segments={props.word.segments}
+      />
+
+      {props.panelType === WordPanel ? (
```
loafofpiecrust, line 123: Hmm it feels weird to use a component as a type distinguisher here, no? Maybe you need an enum?

website/src/edit-word-panel.tsx
chanmioh: Gotcha, went ahead and shortened them to be concise.

website/src/panel-layout.tsx
chanmioh: Since I can't use object literals as values for enums, I have an additional line here now that checks the enum type and then sets the component type that is rendered - I hope this would be okay!

image

website/src/panel-layout.tsx
chanmioh: I reverted the styling for this element, but kept this styling for when the panel contains an edit button if that's okay.

chanmioh:
### CHANGES
Backend
- `update_word.sql`: Simplified SQL query 
- `database_sql.rs`: 
    - `update_word`: Simplified `source` and `commentary` value assignments by abstracting logic into a function implemented in `form.rs`
- `form.rs`: 
    - `id`: Added an `ok_or_else` that throws error if no id exists in an `AnnotatedForm`
    - `FromUndefinedExt`: Defined an “external trait” with a function implementation that checks if given value is undefined, then converts into a vector accordingly

Frontend
- `panel-layout.tsx`: Reverted word panel to original except if user is logged in (edit button will be present, close button the left, etc)
- `edit-word-panel.tsx`: Removed aria-labels since the IconTextButtons contain self-explanatory text
- `form-context.tsx`: 
    - Instantiated `word` variable as an empty object since instantiating it as an `undefined` causes the `onSubmit` in `form` to throw an error if `values.word` is not checked beforehand in `onSubmit` method, even if it is checked in `onValidate` method. Let me know if this is confusing
    - `runUpdate`: Added `async` and `await` 
    - `onValidate`: Checks if values are defined
- `document.page.tsx`: Moved `FormProvider` to `TranslationTab`

Styling
- `panel-layout.css.tsx`: Reverted word panel styling to original, but added two new styles for the editable panel view
- `button.tsx`: Added `children` prop to `IconTextButton` props type

loafofpiecrust (approved):
Looks like you just need to resolve merge conflicts (and also probably rename this trait + method), then you can merge it!

types/src/form.rs
```diff
@@ -169,10 +175,27 @@ pub fn is_root_morpheme(s: &str) -> bool {
...
+}
+
+/// Trait that defines function which takes in a possibly undefined value.
+pub trait FromUndefinedExt<T> {
+    /// If the given value is undefined, convert into a vector of option. Otherwise, return an empty vector.
+    fn from_undefined(self) -> Vec<Option<T>>;
+}
+
+impl<T> FromUndefinedExt<T> for MaybeUndefined<T> {
+    fn from_undefined(self) -> Vec<Option<T>> {
```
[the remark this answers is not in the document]
loafofpiecrust: You could call the trait MaybeUndefinedExt to match the type

website/src/panel-layout.tsx
loafofpiecrust: Yeah that works, though use the enum variant name rather than a number like `0`

types/src/form.rs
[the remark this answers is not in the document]
chanmioh: Makes sense! Done

PR #201, opened 2022-07-15
Added new migration file altering collection_chapter table. Adds new …

nole2701:
Added new migration file altering collection_chapter table. Adds new column indicating intro/body chapter. Currently there is no data in the collection_chapter table. I believe Hazelyn is completing that this week too. So I'm going to have to adjust the data pulling to account for this schema change later.


types/migrations/20220714123910_edited_collections_2.sql
```diff
@@ -0,0 +1,6 @@
+-- Add migration script here
+create type chapter_t as enum ('intro', 'body');
```
loafofpiecrust, line 2: For a `SQL` enum, we need to ensure it works with a corresponding Rust type. Generally that means UpperCamelCase for enum variants. In short, let's do `Intro` and `Body` for these variants too.

types/migrations/20220714123910_edited_collections_2.sql
```diff
@@ -0,0 +1,6 @@
...
+
+alter table collection_chapter
+    add chapter_type chapter_t;
```
loafofpiecrust, line 5: Types are allowed to have the same name as columns, as far as I know. So, we could have `chapter_type` for both. I think we could use a more descriptive name here though since "chapter type" could mean a lot of things, plus we ideally don't need `chapter_` at the beginning of the column name (`collection_chapter.chapter_type` is a bit redundant). Perhaps simply `section`, with type name `collection_section`? What do you think?

types/migrations/20220714123910_edited_collections_2.sql
```diff
@@ -0,0 +1,6 @@
+-- Add migration script here
+create type collection_section as enum ('Intro', 'Body');
+
+alter table collection_chapter
+    add section collection_section;
+
```
loafofpiecrust, line 6: Make sure there's a newline at the end of the file, and would you mind running `sqlfluff fix types/migrations/20220714123910_edited_collections_2.sql` to format it?

types/migrations/20220714123910_edited_collections_2.sql
```diff
@@ -0,0 +1,6 @@
+-- Add migration script here
```
loafofpiecrust, line 1: - Could you rename this migration to something more unique like `edited_collection_intro_chapters` for example?
- You can get rid of this first line, it's a placeholder.

types/migrations/20220714123910_edited_collections_2.sql
```diff
@@ -0,0 +1,6 @@
...
+create type collection_section as enum ('Intro', 'Body');
+
+alter table collection_chapter
+    add section collection_section;
```
loafofpiecrust, line 5: I realized that this is a nullable column, which we don't want. Let's make this `not null` and add `default 'Body'` so that existing rows get the value of `Body`.

loafofpiecrust (commented):
Generally looks good. I noticed you changed the permissions on a few files (`graphql/src/{lambda,query,server}.rs`), mind changing them back?

types/migrations/20220714123910_edited_collections_2.sql
loafofpiecrust, line 1: Looks like you still have the old migration file committed.

loafofpiecrust (approved):
Looks good to merge once you get rid of the old migration file.

PR #204, opened 2022-07-22
[DP-151] Created front end visualization for doc slices in wp

KushP10:
image

allows for functionality in WordPress to cite documents and have words rendered in appropriate webpage

for testing edit [this](https://wp.dailp.northeastern.edu/wp-admin/post.php?post=1130&action=edit) wordpress page

to view changes visit ~/collections/cwkw in browser


website/src/components/wordpress.tsx
```diff
@@ -87,10 +107,81 @@ const parseOptions: HTMLReactParserOptions = {
...
+const PullWords = (props: { slug: string; start: number; end?: number }) => {
+  const slug = props.slug
+  const start = props.start
+
+  var end
+  if (!props.end) {
+    end = start + 1
+  } else {
+    end = props.end
+  }
```
patelShas, line 119: Check if you can condense this to `end = props.end ? props.end : start + 1` or something similar.

website/src/components/wordpress.tsx
```diff
@@ -87,10 +107,81 @@ const parseOptions: HTMLReactParserOptions = {
...
+    useState<Dailp.FormFieldsFragment | null>(null)
+
+  const dialog = useDialogState({ animated: true })
+  const selectAndShowWord = (content: Dailp.FormFieldsFragment | null) => {
+    setSelectedWord(content)
+    if (content) {
+      dialog.show()
+    } else {
+      dialog.hide()
+    }
```
patelShas, line 145: similarly check if this can be condensed.

patelShas (commented):
Checked out a local copy since sometimes the preview doesn't work with backend changes, however on my local copy the website can't run, citing an error in index.ts that leaves it unable to load schema. Perhaps check if there's anywhere in the index.ts where the DocSlice stuff gets referenced but wasn't updated.

website/src/components/wordpress.tsx
```diff
@@ -53,13 +64,22 @@ const parseOptions: HTMLReactParserOptions = {
...
+            ></PullWords>
+          )
         }
         if (segments?.length === 3) {
-          return pullWords(segments[1]!, segments[2]!)
+          return (
+            <PullWords
+              slug={segments[1]!}
+              start={parseInt(segments[2]!)}
+            ></PullWords>
```
patelShas, line 81: I feel like this could be condensed with the previous conditional since everything is the same besides the end argument. However since its only two cases, I think its fine to leave as is if condensing takes too much trouble.

website/src/components/wordpress.tsx
```diff
@@ -87,10 +107,74 @@ const parseOptions: HTMLReactParserOptions = {
...
-  start: string,
-  end?: string
-): JSX.Element => {
-  return <div>full array</div>
+const PullWords = (props: { slug: string; start: number; end?: number }) => {
+  const slug = props.slug
+  const start = props.start
+
+  var end
+  end = props.end ? props.end : start + 1
```
patelShas, line 115: Since end doesn't get assigned again, I think making a single line const would work.

patelShas (commented):
Pretty good, I don't think it needs that much change. Functions well and responsive on front-end. Earlier hiccups with index.ts also are no longer a problem.

website/src/components/wordpress.tsx
KushP10: this clause allows for ease of understandability of the component

loafofpiecrust: @KushP10 I don't know what that means. The comment was just to condense this declaration into a `const` since you don't mutate the variable. Does that seem less readable or something?

website/src/components/wordpress.tsx
```diff
@@ -53,13 +64,22 @@ const parseOptions: HTMLReactParserOptions = {
     if ("data" in node) {
       const segments = node.data.match(style)?.filter((x) => x !== undefined)
       if (segments) {
-        console.log(segments)
-
         if (segments?.length === 4) {
```
loafofpiecrust, line 67: The only difference in these two conditional branches is whether segments[3] is null or not. So I would combine them under a condition like `segments.length >= 3` and check if the last element is null or not when parsing it, something like `segments.length >= 4 ? parseInt(segments[3]) : null`.

loafofpiecrust (approved):
Conditional approval. This is good to go once you address the last comments from @patelShas and me.

website/src/components/wordpress.tsx
KushP10: > @KushP10 I don't know what that means. The comment was just to condense this declaration into a `const` since you don't mutate the variable. Does that seem less readable or something?

Sorry, I misinterpreted what Shashwat's comment meant.

PR #206, opened 2022-08-05
Migrate intro and genre chapters

hazelyn11:
Modified migration process to track intro and body chapters


types/queries/insert_chapter_mark_type_genre.sql
```diff
@@ -0,0 +1,17 @@
...
+  section)
+values (
+  $1,
+  ( select id
+    from document
+    where short_name = $2),
+  $3,
+  $4,
+  $5,
+  'Body'
```
hazelyn11, line 16: I couldn't get passing in a CollectionSection as a sixth parameter to work here: I would get the error "unsupported type collection_section for param #6".

types/src/collection.rs
```diff
@@ -27,4 +28,13 @@ pub struct Chapter {
...
     pub wordpress_id: Option<i64>,
+
+    pub chapter_type: CollectionSection,
+}
+
+#[derive(Debug, Clone, PartialEq, sqlx::Type)]
+#[sqlx(type_name = "collection_section")]
+pub enum CollectionSection {
+    Intro,
+    Body,
```
hazelyn11, line 39: I think we should probably be consistent with calling the second section of chapters either 'Genre' or 'Body' chapters-- if 'Body' chapters is preferred, I can rename other areas where we reference them throughout the migration code. Should I also change the mockup spreadsheet to reflect that?

loafofpiecrust: Yeah I think it should be "Body" because we don't know how future collections will be split up.

loafofpiecrust: And if you could propagate that, that'd be great thanks!

types/queries/insert_chapter_mark_type_genre.sql
loafofpiecrust: Hmm it's should be possible for this to work, did you check out usage of `SegmentType`? If you look at line 758 in database_sql.rs, for that custom enum type we have to tell sqlx to ignore type-checking (which is weird, but seems to work) by adding `as _` to the callsite.

migration/src/spreadsheets.rs
```diff
@@ -171,13 +176,19 @@ impl SheetResult {
                     doc_string = row_values.next();
                 }
 
+                let mut intro_or_genre = Body;
+                if is_intro == true {
+                    intro_or_genre = Intro;
+                }
```
loafofpiecrust, line 182: Aim to avoid `mut` when possible (also doing `== true` with bools), you can do this instead:
```rs
let x = if is_intro { Intro } else { Body };
```

migration/src/spreadsheets.rs
```diff
@@ -171,13 +176,19 @@ impl SheetResult {
...
+
                 let new_chapter = Chapter {
                     index_in_parent: index_i64,
                     url_slug: chapter_url_slug,
                     chapter_name: cur_chapter_name,
                     document_short_name: doc_string,
                     id: None,
                     wordpress_id: wp_id,
+                    chapter_type: intro_or_genre,
```
loafofpiecrust, line 191: Mind matching the new field to the database column name?

hazelyn11: fixed!

types/queries/insert_chapter_mark_type_genre.sql
hazelyn11: adding as _ worked, thank you!

migration/src/spreadsheets.rs
hazelyn11: cool, thanks-- I fixed this in the instance that you pointed out and rewrote a few other lines in parsing the collection index to reflect this.

types/src/collection.rs
hazelyn11: sounds good thanks-- I changed naming across the migration code to reflect this as well as the second section of the spreadsheet mockup.

patelShas (commented):
I think it looks pretty good, without much real change needed. One thing I'm wondering is that since `insert_one_chapter_marking_intro_or_body.sql` is able to do all the functionality of the more specific `insert_chapter_mark_type_intro.sql` and `insert_chapter_mark_type_genre.sql`, those two seem unnecessary, especially since the section arg is basically already specified as an enum in the rust. As is it doesn't have any negative effect, it just seems redundant without any functionality.

types/queries/insert_chapter_mark_type_genre.sql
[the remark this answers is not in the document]
hazelyn11: oh yeah, regarding your point about those two files above, I can just delete them since they're irrelevant now and were from an earlier iteration of the code-- I think that would probably be best?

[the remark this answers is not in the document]
patelShas: Yeah, I think so. And since they're not very complex, we could easily create something similar if we ever needed in the future.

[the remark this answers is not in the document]
hazelyn11: OK, should be all set thanks!

migration/src/spreadsheets.rs
```diff
@@ -178,19 +180,20 @@ impl SheetResult {
...
-        let self_genre_chapters = chapters;
+        let self_body_chapters = chapters;
 
         Ok(Collection {
             title: self_title.to_string(),
             wordpress_menu_id: Some(*self_wordpress_menu_id),
             slug: self_slug.to_string(),
             intro_chapters: self_intro_chapters,
-            genre_chapters: self_genre_chapters,
+            body_chapters: self_body_chapters,
```
loafofpiecrust, line 196: Since we have a marker for this inside the chapter now, do you think it's reasonable to condense this down to one field, `chapters`?

loafofpiecrust (approved):
Looking good! If you don't think it makes sense to, feel free to ignore the change below. But I think it might make your types more reusable.

migration/src/spreadsheets.rs
hazelyn11: I agree, fixed this!

hazelyn11: am I all set to merge now that that's done?

loafofpiecrust: Yup, I already approved it for merging

PR #211, opened 2022-08-09
Fix immediate issues with new display settings

loafofpiecrust:
- Correct misspelling of "syllabary"
- Rename `ViewMode` to `LevelOfDetail`
- Simplify settings code
- Improve accessible labelling of each option


website/src/mode.tsx
```diff
@@ -55,85 +53,69 @@ const cherokeeRepresentationMapping = {
...
-      )}
-    </Select>
-  )
-}
+}) => (
+  <Select
+    id={p.id}
+    name="cherokee-representation"
+    value={p.value}
+    onChange={(e) => p.onSelect(e.target.value as Dailp.CherokeeOrthography)}
```
chanmioh, line 88: I'm not sure if this casting here is needed since the given prop value passed into this component's `value` is already a `Dailp.CherokeeOrthography`.

chanmioh (commented):
Looks great! Only thing I can think of as for design changes would be maybe abstracting `SelectLevelOfDetail` and `SelectCherokeeRepresentation` into a single component since they both contain similar code. I think this would only be necessary though if we were planning to add more preferences in the future of a similar nature. Otherwise looks good!

website/src/mode.tsx
```diff
@@ -55,85 +53,69 @@ const cherokeeRepresentationMapping = {
...
-  )
-}
+export const SelectLevelOfDetail = (p: {
+  id: string
+  value: LevelOfDetail
+  onSelect: (mode: LevelOfDetail) => void
+}) => (
+  <Select
+    id={p.id}
+    name="experience-picker"
```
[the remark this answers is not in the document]
loafofpiecrust: Yup makes sense to me

website/src/preferences-context.tsx
```diff
@@ -15,51 +15,89 @@ export const usePreferences = () => useContext(PreferencesContext)
...
+    return LevelOfDetail.Pronunciation
+  }
+}
+
+const savedCherokeeRepresentation = (): Dailp.CherokeeOrthography => {
+  const newCookie = LocalStorage.getItem(
     PreferenceKey.CherokeeRepresentation
-  ) as Dailp.CherokeeOrthography) ?? Dailp.CherokeeOrthography.Learner
+  ) as Dailp.CherokeeOrthography
+  const oldCookie = Cookies.get(DeprecatedCookie.LevelOfDetail)
```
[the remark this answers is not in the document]
loafofpiecrust: No so actually my recent changes removed the separate distinction of romanization style, folding that into the "Cherokee Representation" option. We can determine the value for that just from the old "Level of detail" cookie, and disregard the "romanization" cookie.

website/src/mode.tsx
loafofpiecrust: Currently the `e.target.value` comes straight from the browser API, where a select value is always a string type. It doesn't adapt its type based on the value type passed in.

website/src/preferences-context.tsx
```diff
@@ -15,51 +15,89 @@ export const usePreferences = () => useContext(PreferencesContext)
...
   )
+
+  useEffect(() => {
+    save(PreferenceKey.LevelOfDetail, levelOfDetail)
+    save(PreferenceKey.CherokeeRepresentation, cherokeeRepresentation)
+    // Remove all deprecated cookies, keep this code for at least several months
+    // to let users clear their site cookies.
+    Cookies.remove(DeprecatedCookie.LevelOfDetail)
+    Cookies.remove(DeprecatedCookie.Romanization)
+  }, [levelOfDetail, cherokeeRepresentation])
```
patelShas, line 30: It seems unnecessary to include an attempt to remove deprecated cookies everytime `levelOfDetail` or `cherokeeRepresentation`, since after its done once we shouldn't really need it done multiple times. I don't know if it involves enough performance to warrant pulling out, but I think it would be sufficient to just pull it into its own `useEffect` that only runs on first render.

website/src/preferences-context.tsx
```diff
@@ -15,51 +15,89 @@ export const usePreferences = () => useContext(PreferencesContext)
...
+  Romanization = "phonetics",
+}
+
+const LocalStorage =
+  typeof window !== "undefined"
+    ? window.localStorage
+    : { getItem: () => null, setItem: () => {} }
+
+const savedLevelOfDetail = (): LevelOfDetail => {
+  const newCookie = LocalStorage.getItem(PreferenceKey.LevelOfDetail)
```
patelShas, line 70: It might be confusing referring to this as a cookie, since a big part of this change is switching over to local storage from cookies. Maybe referring to them as `newData` and `oldData`, or even just `stored` and `cookie`.

website/src/preferences-context.tsx
```diff
@@ -15,51 +15,89 @@ export const usePreferences = () => useContext(PreferencesContext)
...
+  } else if (oldCookie) {
+    const parsed = Number.parseInt(oldCookie) as LevelOfDetail
+    return Math.min(parsed, LevelOfDetail.Segmentation)
+  } else {
+    return LevelOfDetail.Pronunciation
+  }
+}
+
+const savedCherokeeRepresentation = (): Dailp.CherokeeOrthography => {
+  const newCookie = LocalStorage.getItem(
```
patelShas, line 84: Same idea about calling it `newCookie` as before.

patelShas (commented):
Looks good! I don't think much change is need, maybe just some naming and one thing about the cookie removal.

website/src/preferences-context.tsx
loafofpiecrust: Yeah that's fair, will do.

website/src/preferences-context.tsx
loafofpiecrust: Oh yeah, definitely. I wrote the code before using local storage for the new value. Will change.

PR #212, opened 2022-08-12
[DP-168] Added editable glosses for words in word panel

chanmioh:
I wanted to try and keep this PR as small as I could, so for now this PR implements the ability to:
- Select a global gloss for a morpheme
- Create a new local gloss for a morpheme (and I believe with how the `upsert_many_word_segments` SQL query is set up, if the local gloss already exists then it will just find the preexisting the local gloss)

Hoping to add these features in later:
- Modify a morpheme's `morpheme` field from the word panel
- Modify a morpheme's `followedBy` field from the word panel
- Modify one instance of a morpheme in a document versus modify all instances of a morpheme

### Changelog

**Backend**
- `database_sql.rs`: 
    - `update_word`: Makes a call to `upsert_many_word_segments` SQL query to update a given word’s new segmentation
- `form.rs`:
    - `AnnotatedFormUpdate`:
        - Added document id as a field since this is needed to create a new local gloss in `morpheme_gloss`
        - Added segments field 
- `morpheme.rs`:
    - Created `MorphemeSegmentUpdate` as an input object

**Frontend**
- `queries.graphql`: FormFieldsFragment also queries for a document id since this is needed for updating a word
- `document-page.tsx`: `DocumentContents` component will reset the current form if it was not submitted. 
- `form-content.tsx`: 
    - Created variable `updatedSegments` which converts the current segments of a word into `MorphemeSegmentUpdate` type for the backend
    - `runUpdate` now takes in a document id and the updated segments
- `panel-layout.tsx`: 
    - `PanelLayout` component contains backend call with `useGlossaryQuery` to get all the global tags
    - Created `EditSegmentation` component that handles editing a word’s segmentation
    - Created `EditGloss` component that handles editing a segment’s gloss
- `creatable.tsx`: Custom `CreatableSelect` component with custom styling. This component handles the front-end gloss selection on an editable word panel


website/src/edit-word-panel.tsx
```diff
@@ -0,0 +1,100 @@
...
+      ) : (
+        <IconTextButton
+          icon={<HiPencilAlt />}
+          className={css.editPanelButton}
+          onClick={() => {
+            if (cherokeeRepresentation === Dailp.CherokeeOrthography.Taoc) {
+              setIsEditing(true)
+            } else {
+              alert(
+                "Currently, only the linguistic analysis using terms from Tone and Accent in Oklahoma Cherokee (TAOC) is supported for editing. Please update your Cherokee description style in the display settings."
```
patelShas, line 56: I want to point out one edge case. If the user is in TAOC mode, they can open the editing panel. Once open, they are able to change preferences out of TAOC mode, but the edit panel remains open. You should see if you can make the edit panel close when settings are changed, disable saving, or some kind of alert if you try to save.

website/src/pages/documents/document.css.ts
```diff
@@ -194,8 +194,8 @@ export const contentContainer = style({
 
 export const contentSection2 = style({
   display: "none",
-  flex: 1,
```
patelShas, line 197: Fair, I sometimes wonder if we should have done that earlier.

website/src/pages/documents/document.page.tsx
```diff
@@ -239,19 +240,36 @@ const DocumentContents = ({
...
       morphemeSystem: cherokeeRepresentation,
     },
   })
-  const docContents = data?.document
+
+  const docContents = result.data?.document
+  const { form, isEditing } = useForm()
+
+  useEffect(() => {
+    // If the form has been submitted, update the panel's current contents to be the currently selected word
```
patelShas, line 257: I'm not sure if this is working. When I make edits I find that I have to reload the page in order for the changes to appear.

patelShas (commented):
Most of the way there, but I think some changes still needed. Admittedly I'm a bit confused as to what gloss means specifically. 

Changes and edits to translations of word parts do save, and I really like this new menu that appears when you change them. However, there's a text indicator surrounded by a box when you begin making changes. I'd add a screenshot, but I haven't managed to get one while keeping the text cursor. I understand the box can help accessibility, but it especially clashes with when you first make the change, because the old text is still there and they overlay. Either remove the box, or make the text non visible when changing.
I assume that is the part referred to by gloss and not the one labeled English translation. That one does not save, but does update the page by turning it into a white screen. If English translation is included in gloss, well there's an issue. 

The edited changes save but are not reflected until the page is reloaded. If that could be made automatic, I think it would look really good. After that, I think that's the important stuff.

One final note is that if you view the screen on mobile or a small screen, the details panel cannot be scrolled. This isn't strictly what the ticket is about, but since you're already working and changing the panel a bunch, I think you could fix it.

website/src/components/button.tsx
```diff
@@ -17,3 +17,25 @@ export const IconButton = ({
...
+
+type IconTextButtonProps<C extends React.ElementType> = ButtonProps & {
+  icon: ReactNode
+  children: ReactNode
+  as?: C
+}
+
+export const IconTextButton = <C extends React.ElementType>({
+  icon,
+  children,
```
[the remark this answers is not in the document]
chanmioh: That makes sense, and the component is labelled IconTextButton anyways so it would make more sense to have the child be a string.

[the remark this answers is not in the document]
loafofpiecrust: The children prop is almost always the same ReactNode type by convention so I'm less convinced this change is better. We could easily want to use a react fragment inside a button to join multiple strings, or have an inline span tag. I would agree with you for other props though. Nbd we can keep the stricter API until we need to change it.

website/src/pages/documents/document.page.tsx
chanmioh: That's odd.. It's working on my local, but which ones are you editing that don't update? Also, @loafofpiecrust are you having this issue as well?

chanmioh:
> Most of the way there, but I think some changes still needed. Admittedly I'm a bit confused as to what gloss means specifically.
> 
> Changes and edits to translations of word parts do save, and I really like this new menu that appears when you change them. However, there's a text indicator surrounded by a box when you begin making changes. I'd add a screenshot, but I haven't managed to get one while keeping the text cursor. I understand the box can help accessibility, but it especially clashes with when you first make the change, because the old text is still there and they overlay. Either remove the box, or make the text non visible when changing. I assume that is the part referred to by gloss and not the one labeled English translation. That one does not save, but does update the page by turning it into a white screen. If English translation is included in gloss, well there's an issue.
> 
> The edited changes save but are not reflected until the page is reloaded. If that could be made automatic, I think it would look really good. After that, I think that's the important stuff.
> 
> One final note is that if you view the screen on mobile or a small screen, the details panel cannot be scrolled. This isn't strictly what the ticket is about, but since you're already working and changing the panel a bunch, I think you could fix it.

Thanks for the feedback! I went ahead and pushed a new commit to address some of the issues you mentioned. I haven't fixed every issue you mentioned (like getting the details panel to properly scroll) so don't worry too much about checking on the new commit if you're busy!

chanmioh:
### Changelog
- `form-context.tsx`: Checks that the TAOC representation is selected in display settings before submitting “Save” in the edit word panel to prevent an edge case.
- `panel-layout.tsx`: Since saving modified translations seems to throw a white page error, for now I made it so the edited panel does not allow for editing the translations. Can add this support in later!
- `button.tsx`: `IconTextButton` prop takes in a string as a child instead of `ReactNode` in consideration that complex components may not be necessary (TBD).
- `creatable.tsx`:
    - `customStyles` variable directly uses CSS properties since this was more familiar to me.
    - Fixed issue with textbox outline showing up when typing in a gloss.
    - Renamed a few variables.

chanmioh:
Accidentally created new duplicate commits above, feel free to ignore those ones and I'll squash them later.

website/src/pages/documents/document.page.tsx
patelShas: Ah, nvm, that part works.

patelShas (commented):
It all looks really good! I like what you how you went about with all the menus and the edit panel. I haven't found any new edge cases, works with the TAOC change. It also reponds correctly to the same word/morpheme being edited repeatedly, which is quite nice. Great as much as I can tell.

chanmioh:
> It all looks really good! I like what you how you went about with all the menus and the edit panel. I haven't found any new edge cases, works with the TAOC change. It also reponds correctly to the same word/morpheme being edited repeatedly, which is quite nice. Great as much as I can tell.

Sounds good, thanks so much for looking over it! :) Appreciate it a lot.

types/src/database_sql.rs
```diff
@@ -388,10 +390,68 @@ impl Database {
...
+            &*doc_id,
+            &*gloss
+        )
+        .execute(&self.client)
+        .await?;
+
+        query_file!(
+            "queries/upsert_many_word_segments.sql",
+            &*doc_id,
+            &*gloss,
```
loafofpiecrust, line 447: Okay so AFAICT you are using the glosses from the front-end straight-up with no conversion, and those glosses come from TAOC. That's a no-go. The TAOC tags are _not_ the same as the internal ones, though many/most of them appear the same. TAOC is just the main format where we currently have a 1:1 mapping between internal word parts and TAOC word parts (and the one where they're the most similar). So you still need to convert these glosses at some point.

I know it's confusing... Perhaps we can come up with a better scheme at some point.

website/src/form-context.tsx
```diff
@@ -0,0 +1,83 @@
...
+        // Create an array of MorphemeSegmentUpdate type to send to the backend.
+        const updatedSegments: Array<Dailp.MorphemeSegmentUpdate> = values.word[
+          "segments"
+        ].map((segment) => {
+          return {
+            morpheme: segment.morpheme,
+            gloss: segment.gloss,
+            role: segment.role,
+          }
+        })
```
loafofpiecrust, line 53: why is the `.map` needed? `segment` has too many fields, is that right?

You can also do `.map((segment) => ({ morpheme: segment.morpheme, ... }))`

website/src/form-context.tsx
```diff
@@ -0,0 +1,83 @@
...
+            id: values.word["id"],
+            docId: values.word["position"].documentId,
+            source: values.word["source"],
+            commentary: values.word["commentary"],
+            segments: updatedSegments,
+          },
+        })
+      } else {
+        alert(
+          "Currently, only the linguistic analysis using terms from Tone and Accent in Oklahoma Cherokee (TAOC) is supported for editing. Please update your Cherokee description style in the display settings."
```
loafofpiecrust, line 66: Define this string only once somewhere

website/src/edit-word-panel.css.tsx
```diff
@@ -0,0 +1,42 @@
...
+  paddingY(vspace.medium),
+  {
+    display: "flex",
+    justifyContent: "space-around",
+    marginLeft: hspace.small,
+  },
+])
+
+export const cancelButton = style([
+  { fontFamily: fonts.header, width: "4.5rem" },
```
loafofpiecrust, line 25: Use or define a constant to use for the width in `src/style/constants`, please

website/src/edit-word-panel.css.tsx
```diff
@@ -0,0 +1,42 @@
...
+])
+
+export const formInputLabel = style({
+  fontFamily: fonts.header,
+  fontWeight: "normal",
+  fontSize: "0.9rem",
```
loafofpiecrust, line 31: Same here, use a constant or define a new one if we don't have something good.

website/src/form-context.tsx
```diff
@@ -0,0 +1,83 @@
...
+type FormContextType = {
+  form: FormStateReturn<any | undefined>
+  isEditing: boolean
+  setIsEditing: (bool: boolean) => void
+}
+
+const FormContext = createContext<FormContextType>({} as FormContextType)
+
+// Instantiates a form state used to keep track of the current word and information about all its features.
+export const FormProvider = (props: { children: any }) => {
```
loafofpiecrust, line 18: `any` => `ReactNode` from React.

loafofpiecrust (commented):
Here's some comments from me, but looks like you really need to rebase/merge main because you've got stuff here from your last word editing branch and some other mismatches.

types/src/database_sql.rs
loafofpiecrust: Maybe you're using the internal gloss already when you get the TAOC tag in the front-end, so it's all sorted?

website/src/form-context.tsx
chanmioh: Done!

website/src/form-context.tsx
chanmioh: Yes, I used `.map` here since `segment` has too many fields, and I needed to convert it into a `MorphemeSegmentUpdate` object type.

Does this change look right?

image

types/src/database_sql.rs
chanmioh: Yes, you're right in how I'm using the glosses. I'll see if I can convert those TAOC glosses into the internal ones. Would it make sense to do these in the front-end before sending it over into the back-end, or just have the back-end handle it?

graphql/schema.graphql
```diff
@@ -142,20 +142,37 @@ type AnnotatedForm {
...
 input AnnotatedFormUpdate {
 	"""
 	Unique identifier of the form
 	"""
 	id: UUID!
 	"""
-	Original source text that can be either undefined or null
+	The document id where the given form lives
+	"""
+	docId: UUID!
```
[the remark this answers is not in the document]
chanmioh: Makes sense - I had this because I thought it would be easier to send the document id back from the front-end to the back-end so calculating it in the back-end wouldn't be necessary, but that might make it more cluttered.

chanmioh:
### Changelog
Implemented conversion of front-end glosses to the internal ones (which I am assuming come from `abstract_morpheme_tags`).

- `morpheme.rs`, `form-context.tsx`: Modified `MorphemeSegmentUpdate` struct to also contain information of the segment’s system, and made changes to places that utilize this accordingly.
- `find_internal_glosses.sql`: Using a given segment’s gloss and system, its gloss is converted into the internal form.
- `database_sql.rs`: The system-specific glosses are converted with the SQL query.

types/queries/word_by_id.sql
```diff
@@ -0,0 +1 @@
+select * from word where id = $1
```
loafofpiecrust, line 1: Use the existing query for word by id, I think we have one already. If not, then use the BasicWord struct and query for specific columns here. I try to avoid `select *`

types/src/database_sql.rs
```diff
@@ -398,7 +396,76 @@ impl Database {
             "queries/update_word.sql",
             word.id,
             &source as _,
-            &commentary as _
+            &commentary as _,
+        )
+        .execute(&self.client)
```
loafofpiecrust, line 401: Oh all of these queries inside this function need to go into a transaction so that if one fails the whole thing gets rolled back. Make sure to commit or rollback the transaction at any and all return points (since you have an early return in here).

types/src/database_sql.rs
```diff
@@ -398,7 +396,76 @@ impl Database {
...
+        .await?;
+
+        // If word segmentation was not changed, then return early since SQL update queries need to be called.
+        if word.segments.is_undefined() {
+            return Ok(word.id);
+        }
+
+        let document_id = query_file!("queries/word_by_id.sql", word.id)
```
loafofpiecrust, line 409: Couldn't you get this info from the update word query or something? Seems excessive here to have an extra query just to get the document ID, no?

types/src/database_sql.rs
```diff
@@ -398,7 +396,76 @@ impl Database {
...
+            &*gloss
+        )
+        .execute(&self.client)
+        .await?;
+
+        // Then, convert the given glosses if they have an internal for to add to the database.
+        let internal_glosses: Vec<String> = query_file_scalar!(
+            "queries/find_internal_glosses.sql",
+            &*gloss,
+            match system_name {
```
loafofpiecrust, line 453: Whats up with this? You return TAOC no matter what, is the match here in case of supporting other systems in the future? Please add a comment here explaining if you think this makes sense. Also, if you expect to get only TAOC, then perhaps return an error in the else case instead of "TAOC".

website/src/form-context.tsx
```diff
@@ -35,15 +41,30 @@ export const FormProvider = (props: { children: any }) => {
...
-        },
-      })
+      if (cherokeeRepresentation === Dailp.CherokeeOrthography.Taoc) {
+        setIsEditing(false)
+
+        // Create an array of MorphemeSegmentUpdate type to send to the backend.
+        const updatedSegments: Array<Dailp.MorphemeSegmentUpdate> = values.word[
+          "segments"
+        ].map((segment) => ({
+          system: Dailp.CherokeeOrthography.Taoc,
```
loafofpiecrust, line 51: Shouldn't we be able to supply this value from the preferences?

website/src/panel-layout.css.tsx
```diff
@@ -122,10 +124,47 @@ export const editCherHeader = style([
...
+  textOverflow: "ellipsis",
+  whiteSpace: "nowrap",
+  overflow: "hidden",
+})
+
+export const globalGlossTag = style([
+  paddingX(hspace.small),
+  {
+    fontFamily: fonts.header,
+    fontSize: "0.8rem",
```
loafofpiecrust, line 166: Nudge again on using constants for sizes and colors when possible.

website/src/panel-layout.tsx
```diff
@@ -93,6 +130,13 @@ export const PanelLayout = (p: {
 export const PanelContent = (p: {
   panel: PanelType
   word: Dailp.FormFieldsFragment
+  groupedOptions: {
```
loafofpiecrust, line 133: Maybe pull this into its own type definition in case we need to use it elsewhere? Its a bit nested for an inline type definition.

types/src/database_sql.rs
loafofpiecrust: This is the main issue blocking my approval, because I want it to error out of we try to not use TAOC.

types/queries/word_by_id.sql
loafofpiecrust: If you need this query, then only select the field you need and make the name more specific. "document_id_for_word" or some such

website/src/form-context.tsx
chanmioh: I assumed this since only TAOC is supported for updating segments right now, but I can go ahead and change this to use the preferences!

types/src/database_sql.rs
chanmioh: Yeah, the match is there to support other systems in the future, but I was worried if the user somehow managed to get the backend call this function without "TAOC" enabled, then we should handle that case by having the system be TAOC be default since I didn't want there to be an error (in hindsight, makes less sense now I guess) but I can change this to return an error instead.

website/package-lock.json
GracefulLemming, line 1: This file should be restored

GracefulLemming (approved):

PR #215, opened 2022-08-20
Add git pre-commit hook and dev-check command

loafofpiecrust:
- New pre-commit hook enforces 3 rules to start with. (runs in <500ms)
  1. Commits with SQL changes must include changes to `sqlx-data.json`
  2. A new database migration requires updates to database documentation under `doc/database`
  3. GraphQL query changes in `website/` must come with updated TypeScript types
- This hook will prevent you from committing if one of these checks fails. However it is not fool-proof and it's biased toward false negatives, so you may still have problems if it allows you to commit (so double check).
- Adds a new `dev-check` command which does a whole bunch of things for you (takes about 20 seconds to run on my machine)
  - Does all type generation steps for SQL, GraphQL server, and GraphQL client
  - Formats your code
  - Runs type-check for front-end and back-end
- Use `dev-check` when you:
  - Change the GraphQL server at all and want to propagate changes down to the front-end types
  - Have a mysterious mismatch or formatting problem
  - Are about to commit your code
- The GraphQL server now generates a static schema file at `graphql/schema.graphql` that should be committed along with GraphQL server changes. Update it by running the server with `dev-graphql` or running `dev-check`.
- `website` now generates typescript types from the static schema file
- Formats some stray unformatted `.rs` files


loafofpiecrust:
Everyone, are you fine with having a local pre-commit hook with these requirements? Does `dev-check` sound helpful?
I'm hoping this can reduce the burden on code reviewers a little bit by having one catch-all command you can run before committing.

loafofpiecrust:
I'm actually just gonna merge this since I think it's necessary and should be uncontroversial.

PR #220, opened 2022-09-06
[DP-62] As a reader, I want a dedicated paragraph details pane

chanmioh:
### Changelog

- `segment.tsx`: Added a “details button” to the right of paragraphs, similar to the details button to the right of a select-able word.

## Syllabary View

image

## Syllabary, Simple Phonetics & Syllabary, Phonetics, and Word Parts View

image

- `panel-layout.tsx`: Created new type of `FormFieldsFragment` or `TranslatedParagraph` to support selecting paragraphs.
    - `PanelLayout` returns a different layout depending on whether a word or paragraph is selected.
- `paragraph-panel.tsx`: Designed layout of paragraph panel display when a panel is selected.

image

image

- `document.page.tsx`: Now supports selecting a word and a paragraph.
- `wordpress.tsx`: Modified to support selecting a word and a paragraph.


website/src/pages/documents/document.page.tsx
```diff
@@ -128,28 +128,31 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
...
-    setSelectedWord(content)
     if (content) {
       dialog.show()
     } else {
       dialog.hide()
     }
   }
 
-  // When the mobile version of the word panel is closed, remove any word selection.
+  // When the mobile version of the word panel is closed, remove any segment selection.
```
hazelyn11, line 145: Is segment here just referring to a paragraph? I know in other areas of the code, segment refers to a line and paragraphs are called blocks. I could also be misinterpreting what's going on here though

hazelyn11: Oh, actually, after looking a bit closer, it looks like now either a word or a paragraph could be considered a segment, which makes more sense!

website/src/segment.tsx
```diff
@@ -106,7 +142,9 @@ export const AnnotatedForm = (
   if (showAnything) {
     const showSegments = p.levelOfDetail >= LevelOfDetail.Segmentation
     const translation = p.segment.englishGloss.join(", ")
-    const isSelected = p.wordPanelDetails.currContents?.id === p.segment.id
+    const isSelected =
+      (p.wordPanelDetails.currContents as Dailp.FormFieldsFragment)?.id ===
+      p.segment.id
```
hazelyn11, line 147: Overall, I felt I could pretty easily follow your code despite the fact that I'm not familiar with this area of the project. It took me a second to realize that a segment refers to either a word or a paragraph, but you make this pretty clear with comments, and you're very consistent with naming things throughout.

website/src/pages/documents/document.page.tsx
```diff
@@ -128,28 +128,31 @@ const TranslationTab = ({ doc }: { doc: Document }) => {
...
     if (!dialog.visible) {
-      selectAndShowWord(null)
+      selectAndShowContent(null)
     }
   }, [dialog.visible])
 
-  let wordPanelInfo = {
-    currContents: selectedWord,
-    setCurrContents: selectAndShowWord,
+  // This can now be either a word or paragraph.
```
hazelyn11, line 152: Before merging, maybe just take "now" out of this comment since someone looking at the updated code won't see the old code

website/src/paragraph-panel.tsx
```diff
@@ -0,0 +1,57 @@
...
+        aria-label="Dismiss selected paragraph information"
+      >
+        <MdClose size={32} />
+      </IconButton>
+
+      <header className={css.wordPanelHeader}>
+        <h1 className={css.noSpaceBelow}>
+          Selected
+          <br />
+          Paragraph
```
loafofpiecrust, line 32: Do you think we could show the paragraph number in the header? Same with the selected word. Instead of "selected word", "word #x"

loafofpiecrust (commented):
Do you think putting the selection icon centered to the right of the paragraph would be a good experience? Semantically where you put it makes sense since right now only the translations will be editable in the panel. But it does look perhaps a little awkward and easy to miss. Open to other ideas for placement.

website/src/pages/documents/document.page.tsx
chanmioh: Makes sense!

website/src/pages/documents/document.page.tsx
chanmioh: Yeah, I think throughout the code, the meaning of segment is pretty flexible. Glad to know it's not too confusing!

chanmioh:
### Changelog
Modified some comments and the `DocumentParagraph` struct now contains an `index` field that represents a 1-indexed position of a paragraph within its document.

- `document.rs`: Added `index` field and its resolver
- `database_sql.rs`: Used enumerate() function to get the index of an object in an iterable, and assigned this as the index of a `DocumentParagraph`
- `queries.graphql`: Added `index` field to `DocumentContents` query
- `panel-layout.tsx` and `paragraph-panel.tsx`: Added index number to the panel header (see below)
image

image

- `edit-word-panel.tsx`: Removed `form.reset()` since this causes form to clear if user "cancels" but enters "edit" mode again.

chanmioh:
> Do you think putting the selection icon centered to the right of the paragraph would be a good experience? Semantically where you put it makes sense since right now only the translations will be editable in the panel. But it does look perhaps a little awkward and easy to miss. Open to other ideas for placement.

I'll have to give this some more thought - but I do agree the selection icon is a bit easy to miss at the moment.

loafofpiecrust (approved):

PR #225, opened 2022-09-19
[DP-104] Added 2 resolvers to query collection structure

nole2701:
After opting to git push --force github says the branch can just merge into the main through this PR. Merging before pushing would leave the branch messed up due to a bad accidental push I did earlier so I decided that pushing then merging would make things less complicated.

I left the comments in the query files.


graphql/src/query.rs
```diff
@@ -18,6 +18,18 @@ pub struct Query;
 
 #[async_graphql::Object]
 impl Query {
+    // query for 1 collection based on slug, and make a collection object with all the stuff in it.
+    async fn get_edited_collection(
```
loafofpiecrust, line 22: Let's call this `edited_collection`, we always use noun phrases for GraphQL fields (rather than verbs like with normal functions).

graphql/src/query.rs
```diff
@@ -18,6 +18,18 @@ pub struct Query;
...
+        &self,
+        context: &Context<'_>,
+        slug: String,
+    ) -> FieldResult<Option<Vec<EditedCollection>>> {
```
loafofpiecrust, line 26: The return type should be singular, right? You're querying for a single collection, not several of them. Something like `FieldResult<Option<EditedCollection>>`

types/src/edited_collection.rs
```diff
@@ -0,0 +1,47 @@
...
+use {
+    crate::async_graphql::{self, dataloader::DataLoader, Context, FieldResult},
+    crate::Database,
+};
+
+/// Structure to represent an edited collection. Missing certain fields and chapters in it.
+/// Used for sending data to the front end
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+#[graphql(complex)]
+pub struct EditedCollection {
```
loafofpiecrust, line 13: Move these structs into the `collection.rs` file. Then put the ones used for data ingestion inside a `raw` submodule. Something like this:

```rs
struct EditedCollection { ... };
struct CollectionChapter { ... };

enum CollectionSection { ... };

mod raw {
    struct EditedCollection { ... };
    struct CollectionChapter { ... };
}
```

Match the `raw::` names like that so it's clear that they're semantically related. Then for usages we'll either see `EditedCollection` or `raw::EditedCollection`.

types/src/edited_collection.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub wordpress_id: std::option::Option<i64>,
+    /// Order within the parent chapter or collection
+    pub index_in_parent: i64,
+    /// Whether the chapter is an "Intro" or "Body" chapter
+    pub section: CollectionSection,
+}
+
+#[async_graphql::ComplexObject]
+impl EditedCollection {
+    async fn all_chapters(&self, context: &Context<'_>) -> FieldResult<Option<Vec<ChapterSingle>>> {
```
loafofpiecrust, line 41: Rename this to `chapters`, we don't need `all_` at this level since there aren't multiple resolvers returning different chapters here.

types/src/database_sql.rs
```diff
@@ -1555,6 +1558,74 @@ impl From<BasicWord> for AnnotatedForm {
     }
 }
 
+// Loader to get each chapter id from the database regardless if it's a subchapter or not.
+#[async_trait]
+impl Loader<CollectionChapter> for Database {
```
loafofpiecrust, line 1563: Name your ID type more specifically based on the operation you're performing, ideally with the name of the result type coming first. Instead of `CollectionChapter` as you have, which is just the type of return value, I would go for something like `ChaptersInCollection`.

types/src/database_sql.rs
```diff
@@ -1555,6 +1558,74 @@ impl From<BasicWord> for AnnotatedForm {
...
+                    ChapterSingle {
+                        id: chapter.id,
+                        title: chapter.title,
+                        wordpress_id: chapter.wordpress_id,
+                        index_in_parent: chapter.index_in_parent,
+                        section: if chapter.section == Some("Intro".to_string()) {
+                            CollectionSection::Intro
+                        } else {
+                            CollectionSection::Body
+                        },
```
loafofpiecrust, line 1591: So this isn't the worst way to do it, but we want to be more future proof. The string could change, or we might add new variants. We want this code not to change when we do that. Look at the `impl Loader<PartsOfWord>` and the `word_parts.sql` file for an example where we get sqlx to convert the type for us. Relevant line:

```sql
  word_segment.role as "role: WordSegmentRole"
```

For you, it might be:

```sql
  section as "section: CollectionSection"
```

This tells sqlx that we want this `text` column in the `collection_section` domain to be parsed into the `CollectionSection` rust type.

types/src/database_sql.rs
```diff
@@ -1555,6 +1558,74 @@ impl From<BasicWord> for AnnotatedForm {
...
+                    },
+                )
+            })
+            .into_group_map())
+    }
+}
+
+#[async_trait]
+impl Loader<CollectionObject> for Database {
```
loafofpiecrust, line 1600: Again, call your key type something more specific based on the operation. Perhaps, `EditedCollectionDetails` or `EditedCollectionId` which would be similar to other keys we have.

types/src/database_sql.rs
```diff
@@ -1555,6 +1558,74 @@ impl From<BasicWord> for AnnotatedForm {
...
+    type Value = Vec<EditedCollection>;
```
loafofpiecrust, line 1601: I see, your output should be singular `EditedCollection`.

types/src/database_sql.rs
```diff
@@ -1555,6 +1558,74 @@ impl From<BasicWord> for AnnotatedForm {
...
+                    CollectionObject(collection.slug.clone()),
+                    EditedCollection {
+                        id: collection.id,
+                        title: collection.title,
+                        wordpress_menu_id: collection.wordpress_menu_id,
+                        slug: collection.slug,
+                    },
+                )
+            })
+            .into_group_map())
```
loafofpiecrust, line 1625: Look at other Loader impls that have singular result types, they use a different method than `.into_group_map()`, since you need to assume a single value per key rather than grouping values with the same key.

loafofpiecrust (changes requested):

types/src/collection.rs
```diff
@@ -1,37 +1,93 @@
...
+#[sqlx(type_name = "collection_section")]
+pub enum CollectionSection {
+    /// Intro chapter
+    Intro,
+    /// Body chapter
+    Body,
+}
 
-    pub url_slug: String,
+pub mod raw {
```
loafofpiecrust, line 49: Uh okay so thinking about this further, making a `pub mod raw` here is limiting because then other modules can't expose raw structs. So if we're gonna have a public `raw` module it should actually be a new folder with mirrored structure, so `types/src/collection.rs` and `types/src/raw/collection.rs` so that we could have other things inside that `raw` folder. Does that make sense to you?

loafofpiecrust (approved):
Looks good once you properly resolve the merge conflicts.

migration/src/edited_collection.rs
```diff
@@ -1,7 +1,7 @@
 use crate::spreadsheets::SheetResult;
 use anyhow::Result;
-use dailp::collection::raw::CollectionChapter;
-use dailp::collection::raw::EditedCollection;
+use dailp::raw::raw::CollectionChapter;
+use dailp::raw::raw::EditedCollection;
```
loafofpiecrust, line 4: There shouldn't be a double `raw` here right?

types/src/raw/collection.rs
```diff
@@ -0,0 +1,42 @@
+use std::option::Option;
+use uuid::Uuid;
+use {
+    crate::async_graphql::{self, dataloader::DataLoader, Context, FieldResult},
+    crate::Database,
+};
+
+pub mod raw {
```
loafofpiecrust, line 8: Take out the extra `raw` module here

PR #226, opened 2022-09-26
[DP-89] As a learner, I want to be able to print a lesson

chanmioh:
### Changelog
Created a stylesheet called `print.css` which formats a lesson page. Custom classes/styles were added. The wordpress code needed to be modified for the styling to work (to create separation of elements when formatting) so for now I've modified the lesson three/Communications to work with this code: https://wp.dailp.northeastern.edu/wp-admin/post.php?post=673&action=edit

NOTE: I wasn't able to add the WP bracket encodings yet since they don't seem to work on my local environment, so it wasn't possible to test for me.

**Created CSS Classes**
`section`: Represents an element that should not be broken up between pages. (i.e. sections with headings)

image

image

`prose`: Represents an element with some kind of content that needs the full-width of the page.

image

Linked file to view the formatted print pdf:
[DAILP.pdf](https://github.com/NEU-DSG/dailp-encoding/files/9649791/DAILP.pdf)


GracefulLemming (changes requested):
The output looks solid! One (admittedly large) question: Is there a reason you did not make this a typescript stylesheet (*.css.ts)? It would be more consistent with the rest of our styles at the very least.

loafofpiecrust:
The styles should definitely be `.css.ts` instead of plain CSS to keep consistent with the rest of our styles. We have some global rules already which you can model off of, and all of the pseudo selectors you're using should be usable there too.

loafofpiecrust:
@chanmioh Could you give more detail on changes you made on the WP page?

loafofpiecrust:
Perhaps the input boxes should have a black outline rather than gray, since gray doesn't always print well. The crafted print PDFs seem to have black borders for those.

chanmioh:
### Changelog
Created a `print.css.ts` file that contains the previous `print.css` styles in TypeScript. 

`print.css.ts`: Uses `globalStyles` to style elements defined under an element with the class `wp`.
`wordpress.tsx`: `parseOptions` looks for an element with a class called `print-lesson` (which is defined in the Wordpress page code) to style it with the vanilla-extract generated CSS class `wp`.
`global.css.ts`: Added some global print styles to certain elements since I assumed they should be the same for all printed pages.
     - Buttons (and input type with radio, since I assume there's no use for that in a print) are hidden.
     - Inputs are styled as a single bottom black line.


[DAILP.pdf](https://github.com/NEU-DSG/dailp-encoding/files/9678102/DAILP.pdf)

chanmioh:
> @chanmioh Could you give more detail on changes you made on the WP page?

Yep!

### Wordpress Changes
All changes can be viewed in: https://wp.dailp.northeastern.edu/wp-admin/post.php?post=673&action=edit

- `print-lesson`: This class was added to the first div element to distinguish WP pages that need to be printed with the lesson stylesheet.
- `section`: This class prevents elements from being broken up between pages. This was used for creating pagebreaks for each new `h1` (i.e. 1. Introduction, 2. Reading ...) and for keeping sections of the lesson page together such as related lists of questions, or vocabulary. 
image
- `prose`: Some content requires being full-width such as the introductory prose, and I saw this content was being broken up by `<br/>`. This class styles certain elements to hide this, since some elements require it for formatting purposes.
image
- `syllabary-input`: Some input fields should be numbered with large textboxes rather than just the normal black lines, so I created this class to account for those fields. There was also a `show-on-print` class included in this formatting since the numbered textboxes shouldn't show on the website, but on the print.
image

image

- `short-answer`: This was styled so the input fields take up the rest of the space of a line with the question, rather than starting on the next line.
image

image

- `centered`: A class used to style the rows of a `table` element, since I needed the image and its blue text centered in a row.
image

image

loafofpiecrust:
> wordpress.tsx: parseOptions looks for an element with a class called print-lesson (which is defined in the Wordpress page code) to style it with the vanilla-extract generated CSS class wp.

Would it make sense to instead check the URL we're rendering or apply to all WP pages? It's okay if you think this is the best solution, I'm just trying to reduce the amount of special handling we need to do on the WP end.

loafofpiecrust:
> * `prose`: Some content requires being full-width such as the introductory prose, and I saw this content was being broken up by `<br/>`. This class styles only certain elements to hide this, since some elements require it for formatting purposes.

Oh! This one is weird. Ideally, the WP paragraphs shouldn't have `<br/>` tags inside of it and should flow with the page. This should be an editing artifact you can get rid of by removing newlines in the WP page editor. Does that sound about right to you?

website/src/components/print-lesson.css.ts
```diff
@@ -0,0 +1,144 @@
...
+
+//// PROSE & SECTIONS
+// Prose is an element that needs to take up the full-width of a page.
+const prose = `${wp} .prose`
+// Section is an element that should not be broken up between pages.
+const section = `${wp} .section`
+
+// This removes line breaks to create full-width pages.
+globalStyle(`${prose} br`, {
+  ...media(mediaQueries.print, {
```
loafofpiecrust, line 39: Does the `globalStyle` function stop you from doing `globalStyle(..., media(mediaQueries.print, ...))` directly?

website/src/components/wordpress.tsx
```diff
@@ -78,6 +81,12 @@ const parseOptions: HTMLReactParserOptions = {
           props["href"] = props["href"].slice(wordpressUrl.length)
         }
         return <Link {...props}>{domToReact(node.children, parseOptions)}</Link>
+      } else if (node.attribs["class"]?.includes("print-lesson")) {
+        return (
+          <div className={wpCss.wp}>
```
loafofpiecrust, line 86: this css style class probably deserves a more specific name, since it's not for just any WP content so far.

chanmioh: sounds good, renaming this to `lesson`!

website/src/components/print-lesson.css.ts
chanmioh: yeah, I have to do `...media()` anytime I want to use that function. since the `...media()` makes the code look a bit cluttered, I can create a function within this file that does the `...media()` call which the globalStyles will call instead, which may look cleaner.

chanmioh: oh actually I can do `globalStyle(..., media(mediaQueries.print, ...))` just fine, I just had to get rid of the `{}` between `...media()`. I can still go ahead and just create a function since there's a lot of duplicate code involving `mediaQueries.print`

chanmioh:
> > * `section`: This class prevents elements from being broken up between pages. This was used for creating pagebreaks for each new section of content (i.e. 1. Introduction, 2. Reading ...) and for keeping sections of the lesson page together such as related lists of questions, or vocabulary.
> 
> Hmmm, it looks like all of the top-level elements have `section`, is that right? If so, I think you can do `.print-lesson > *` (all direct descendants) instead of `.print-lesson .section` (all `.section` descendants recursive) for the selector.

The `section` was also used for some descendants of top-level elements with `section`, here's an example: 

image

This allows for keeping together parts of the lesson that aren't at the top-level, but still needs to start on a new page if not all of its content can fit on one page.

chanmioh:
> > * `prose`: Some content requires being full-width such as the introductory prose, and I saw this content was being broken up by `<br/>`. This class styles only certain elements to hide this, since some elements require it for formatting purposes.
> 
> Oh! This one is weird. Ideally, the WP paragraphs shouldn't have `<br/>` tags inside of it and should flow with the page on screen and print. This should be an editing artifact you can get rid of by removing newlines in the WP page editor. Does that sound about right to you?

Gotcha, I was wondering if the `<br/>` was intentional, so I didn't want to mess with that in the WP page. I can go ahead and remove those new lines though and get rid of this CSS class!

loafofpiecrust:
> Gotcha, I was wondering if the `<br/>` was intentional, so I didn't want to mess with that in the WP page.

It's hard to get nice HTML using the visual editor in WP 😦

loafofpiecrust:
@chanmioh If any of these suggestions seem like a bigger task, we can put them off for now. Just say the word.

chanmioh:
> > * `centered`: A class used to style the rows of a `table` element, since I needed the image and its blue text centered in a row.
> 
> Hmm these aren't semantic tables, so they shouldn't really be table elements ideally. They should be [`fig*` elements](https://www.w3schools.com/tags/tag_figcaption.asp), perhaps if we use those elements that can replace the css class?

Yeah that makes sense - I'll go ahead and change those to be figure elements!

chanmioh:
### Changelog

- `WP page`:
     - Removed `prose` element.
     - Removed unnecessary linebreaks between sentences that should appear both on-screen and print media.
     - Replaced `<table>` element when utilized for image and caption pairs with a `<figure>` element. Styled the figure element with flexbox accordingly. It should look very similar to how it did as a table:
image

     - Added a `<hr>` element underneath the `<figure>` element
image

- `global.css.ts`: `<hr>` styled to not appear on print.
- `wordpress.tsx`: Checked for URL path before parsing to see if the new print stylesheet needs to be included.
- `print-lesson.css.ts`: 
     - Added function `printMedia` to minimize duplicate code of specifying the media query.
     - Removed styles for `<table>`.
     - Added styled for `<figure>`.

GracefulLemming (approved):
With the changes made, this looks really nice! @chanmioh, we should follow-up about documenting the WP changes in the wiki so our content team can reference it!

chanmioh:
> With the changes made, this looks really nice! @chanmioh, we should follow-up about documenting the WP changes in the wiki so our content team can reference it!

Sounds good! I'll go ahead and merge this in for now.

PR #227, opened 2022-09-28
Add slug and collection_slug columns to collection_chapter

loafofpiecrust:
Add these two generated columns to collection_chapter to make certain queries more convenient. Also, add a constraint that requires chapter slugs be unique within each collection.

### DIFF
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -8,7 +8,7 @@ git-staged() {
 }
 
 # Require sqlx data to be generated when a SQL query is modified.
-if git-staged "types/**.sql" && ! git-staged "sqlx-data.json"; then
+if git-staged "types/queries/**.sql" && ! git-staged "sqlx-data.json"; then
     printf "\nDatabase queries changed without generating sqlx types.\nPlease run 'dev-check' or 'dev-generate-types'.\n"
     HOOKS_FAILED=1
 fi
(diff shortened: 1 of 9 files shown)


chanmioh (approved):
I think this looks good! I guess the only question I have is what we'd need the `chapter_path` for, since we're using a chapter's `collection_slug` and `slug` to reference a chapter.

loafofpiecrust:
The `chapter_path` will define the hierarchy, which we'll use for table of contents and breadcrumbs

nole2701:
> The `chapter_path` will define the hierarchy, which we'll use for table of contents and breadcrumbs

Would keeping chapter_path also help to keep old usages working without needing to go back and change them all?

loafofpiecrust:
The reason to have this setup instead of directly storing `slug` and a `parent_id` is that with an ltree we don't have to do a recursive query to figure out the full path of a chapter, which will be a common operation.

loafofpiecrust:
> > The `chapter_path` will define the hierarchy, which we'll use for table of contents and breadcrumbs
> 
> Would keeping chapter_path also help to keep old usages working without needing to go back and change them all?

Yes I suppose, but the alternative of getting rid of `chapter_path` would require a restructuring of the table in general. These new fields are generated from `chapter_path`, we can treat them like functions. You can't change their value directly, you can only observe the value after changing `chapter_path`.

nole2701 (approved):
It looks good. With the additions relating to chapter_path and slugs if they're derived from chapter_path then it should be fine.

PR #229, opened 2022-10-06
[DP-180] Fixed mobile display of lesson pages

chanmioh:
# Changelog
Original bug: On mobile, some elements on the WP page cause the page to overflow, so the whole page will look like this:

image

To fix this, I added mobile query styling to these elements. I didn't want to mess with the WP page elements like blockquote's inline styles too much, so instead of using `min-width` I use `max-width` to override inline styles on the mobile sizes. Lmk if this should be done another way, but that's how I approached this. 

- `wordpress.tsx`: Checked for URL path before parsing to see if the created stylesheet needs to be included.
- `lesson.css.ts`: Added styles for elements specifically on mobile screen sizes. I also added some styles applied to all screen sizes since I thought they'd look a bit better, I elaborate those below.

## WP Changes
Modified WP page: https://wp.dailp.northeastern.edu/wp-admin/post.php?post=673&action=edit

- Created new class called `fill-blank` that styles questions where the input box appears within a sentence, and also contains a word below it. I replaced the previous way it was styled into a flexbox for both mobile AND desktop, so that it would be responsive for both sizes.

Before and after on mobile:
image image

- Created new class called `module-buttons` that repositions the buttons at the end of the lesson. It was positioned awkwardly on mobile, so I turned it into a flexbox for responsiveness on mobile + desktop.

Before and after on mobile:
image image


website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+  thickness,
+  vspace,
+} from "src/style/constants"
+import { marginY, paddingX } from "src/style/utils"
+
+// Parent class defining a Wordpress page.
+// All classes defined in this file begin with this generated className to control the scope of these styles.
+export const lesson = style({})
+
+const mobileQuery = "screen and (max-width: 52em)"
```
loafofpiecrust, line 18: Add this to our media query constants as `smallOnly`

website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+
+//// INPUT STYLES
+const mobileInputSize = `calc(1rem * 2)`
```
loafofpiecrust, line 21: Can't this just be `2rem`?

website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+      width: hsize.full,
+      height: mobileInputSize,
+    }),
+  },
+})
+
+// A global style that styles fill in the blank sentences where a word
+// appears below the input box.
+globalStyle(`${lesson} .fill-blank span`, {
+  display: "inline-flex",
```
loafofpiecrust, line 36: What does `inline-flex` do? Why not normal `flex`?

website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+
+//// IMAGE AND CAPTION STYLES
+const figure = `${lesson} figure`
+// Gives flexbox items (image and caption) a column look only on mobile.
+globalStyle(figure, {
+  display: "flex",
+  flexDirection: "row",
+  alignItems: "center",
+  textAlign: "left",
+  gap: `calc(${space.large} * 2)`,
```
loafofpiecrust, line 50: Add a constant here if you need it, perhaps `space.xlarge` or `space.xxlarge` or something.

website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+// Styles rows of a table.
+// On mobile, since row width is limited, convert each row into a "column-like" element
+// and space out their margins.
+globalStyle(`${lesson} table tr`, {
+  "@media": {
+    [mobileQuery]: important({
+      display: "flex",
+      flexDirection: "column",
+      border: `${thickness.thin} solid ${colors.text}`,
+      ...(marginY(vspace.large) as Styles),
```
loafofpiecrust, line 78: I'm surprised this `as` clause is necessary, it wasn't passing typecheck? Later on, `padding` seems to work fine so maybe revise the type of the `marginY` function.

website/src/components/lesson.css.ts
```diff
@@ -0,0 +1,117 @@
...
+globalStyle(
+  `${lesson} hr`,
+  important({ width: hsize.full, marginBottom: vspace["1.5"] })
+)
+
+/// BLOCKQUOTES STYLES
+// By default, set this width to 500px, but on mobile make width auto to
+// compensate for the smaller size screen.
+globalStyle(`${lesson} blockquote`, {
+  width: `calc(${hsize.large} / 2)`,
```
loafofpiecrust, line 104: If possible, again, add this as a constant to `hsize` or `size` because we might want to use a similar size again.

website/src/components/wordpress.tsx
```diff
@@ -44,8 +46,19 @@ export const WordpressPageContents = ({
 }: {
   content: string | null
 }) => {
+  const { "*": slug } = useRouteParams()
+
+  let parsed
+
   if (content) {
-    const parsed = parse(content, parseOptions)
```
loafofpiecrust, line 48: You can move `let parsed` into the `if` block. Plus it looks like you run the same `parse(content, parseOptions)` expression in both inner blocks. So you could do

```ts
const parsed = parse(content, parseOptions)
if (slug?.includes("lessons/")) {
  return <div ...>{parsed}</div>
} else {
  return parsed
}
```

website/src/components/lesson.css.ts
chanmioh: Do you mean revise the type that `marginY` returns? Since it looks like the `padding` functions that work return a `Styles` instead of `StyleRules` which allows them to work

LN7-cyber: Let's try this out chanmi, and deploy changes after @GracefulLemming reviews.

PR #232, opened 2022-10-26
Update toc

hazelyn11:
Update the table of contents code to support third collection_section type (Credit)

### DIFF
--- a/migration/src/spreadsheets.rs
+++ b/migration/src/spreadsheets.rs
@@ -8,6 +8,7 @@ use anyhow::Result;
 use dailp::collection::CollectionSection;
 use dailp::collection::CollectionSection::Body;
 use dailp::collection::CollectionSection::Intro;
+use dailp::collection::CollectionSection::Credit;
 use dailp::raw::CollectionChapter;
 use dailp::raw::EditedCollection;
 use dailp::{
@@ -145,10 +146,11 @@ impl SheetResult {
         let second_value = row
             .next()
             .ok_or_else(|| anyhow::format_err!("Missing second value"))?;
-        let mut is_intro = true;
+        // 0 for Intro, 1 for Body, 2 for Credit
+        let mut chapter_type = 0;
         for cur_row in row {
             if cur_row[0].is_empty() {
-                is_intro = false;
+                chapter_type = chapter_type + 1;
             } else {
                 let mut row_values = cur_row.into_iter().peekable();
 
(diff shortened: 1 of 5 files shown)


migration/src/spreadsheets.rs
```diff
@@ -176,7 +178,9 @@ impl SheetResult {
                     None
                 };
 
-                let intro_or_body = if is_intro { Intro } else { Body };
+                let intro_or_body = if chapter_type==0 { Intro }
```
chanmioh, line 181: Slight nit, but would just space it out so it looks like `chapter_type == 0` for readability!

chanmioh: And I would maybe rename the variable `intro_or_body` to `chapter_type_name` since it can also be a `Credit` and the variable name could be misleading

migration/src/spreadsheets.rs
```diff
@@ -145,10 +146,11 @@ impl SheetResult {
         let second_value = row
             .next()
             .ok_or_else(|| anyhow::format_err!("Missing second value"))?;
-        let mut is_intro = true;
+        // 0 for Intro, 1 for Body, 2 for Credit
+        let mut chapter_type = 0;
         for cur_row in row {
             if cur_row[0].is_empty() {
-                is_intro = false;
+                chapter_type = chapter_type + 1;
```
chanmioh, line 153: If I'm understanding this code right, when `cur_row[0]` is empty, then that defines the start of a new chapter type? If I'm not understanding that correctly, is it possible for `chapter_type` to ever exceed 2 and cause issues?

chanmioh (commented):
Looks good! I just left a few comments, but they were mainly questions about the code.

migration/src/spreadsheets.rs
hazelyn11: good catch, thanks!!

migration/src/spreadsheets.rs
hazelyn11: cur_row[0] represents the Chapter Depth column in the spreadsheet (https://docs.google.com/spreadsheets/d/12R07Ks8A5g2jffqoILJw7nRn6GehwNwdwosxaJBwNNM/edit?usp=sharing). Each chapter type is separated by the yellow highlighted row, and Chapter Depth is required for every chapter so yes, it should only ever be empty when a new type is being defined. As the spreadsheet is formatted right now I do not believe it will ever be possible for chapter_type to exceed 2, but I thought this strategy might make it easier to add new chapter types and handle that during ingestion if we ever need to

hazelyn11: ^^ let me know if this makes sense or if you still have questions!! thanks for your comments! just committed again with the fixes you suggested

chanmioh: yes that makes sense! Sounds good to me!

chanmioh (approved):
Looks good to me!

PR #234, opened 2022-10-28
[DP-85 + 119] Download Audio + Scroll to top button

nole2701:
Added download button. Did not end up using separate function but found a way to use href on the button itself.


website/src/pages/documents/document.page.tsx
```diff
@@ -347,7 +355,13 @@ export const DocumentTitleHeader = (p: {
           audioUrl={p.doc.audioRecording.resourceUrl}
           showProgress
         />
+        {p.doc.audioRecording && (
+        <div>
+        {!isMobile ? <a href={p.doc.audioRecording?.resourceUrl} download={"audio"}><Button>Download Audio</Button></a> : null}
```
chanmioh, line 360: To avoid the CORS error, are there any links to the audio file that aren't the `resourceUrl`? Looks like you can set a `download` attribute on files that don't link to a different domain: https://stackoverflow.com/a/64719808

chanmioh, line 360: Also the `download` attribute should contain the name of the document containing the audio recording, not just "audio". I think you have access to this field in the `doc` prop.

website/src/pages/documents/document.page.tsx
```diff
@@ -82,6 +83,10 @@ const TabSet = ({ doc }: { doc: Document }) => {
           </Tab>
         </TabList>
       </div>
+      <div id="scroll-top" className={css.scrollTop} onClick={() => window.scrollTo({top: 0, behavior: 'smooth'})}>
```
GracefulLemming, line 86: Maybe consider using an anchor tag somewhere in the scroll to top button so that hovering over the element indicates someone can interact with it. This also helps browsers and screen readers understand what that element does. On that note, this element seems like it is not easily accessible through non-pointer navigation. Maybe consider using tabbable or sim.?

GracefulLemming: You should definitely have some text like "scroll to top" alongside the arrow. Adding a text label makes the visual semantics more clear. As it stands now, the function of the arrow is unclear. On mobile (and maybe even on desktop), adding just "Top" or "To top" should be enough. We could ask Ellen about using ᎤᏍᎪᎵ for 'top, peak' in addition.

website/src/pages/documents/document.page.tsx
```diff
@@ -82,6 +83,10 @@ const TabSet = ({ doc }: { doc: Document }) => {
...
+      <h3><RiArrowUpCircleFill size={55}/></h3>
```
GracefulLemming, line 87: _Please, keep your HTML semantic!_ [Quoting MDN](https://developer.mozilla.org/en-US/docs/Glossary/Semantics): "HTML should be coded to represent the data that will be populated and not based on its default presentation styling." In this case, that means you should **not** use a heading element here since this button does not function as a heading. An anchor link could work, or maybe a non-semantic element like div.

website/src/pages/documents/document.page.tsx
```diff
@@ -331,12 +336,20 @@ export const DocumentTitleHeader = (p: {
...
+
     <div className={css.bottomPadded}>
       {p.showDetails ? (
         <Link href={documentDetailsRoute(p.doc.slug!)}>View Details</Link>
       ) : (
         <Link href={documentRoute(p.doc.slug!)}>View Contents</Link>
       )}
+      {!p.doc.audioRecording && (
+        <div id="no-audio-message">
+        {!isMobile ? <a><b>No Audio Available</b></a> : null}
```
GracefulLemming, line 350: This might be a nitpick, but please use the more semantically appropriate strong element here in place of the bring attention b element.

GracefulLemming (changes requested):
Generally this seems logically reasonable, but there are a number of areas where you should think more about accessibility (tabbability, semantic HTML) and user experience (how the button reacts to user interaction).

website/src/pages/documents/document.css.ts
```diff
@@ -224,3 +225,18 @@ export const mobileWordPanel = style([
     },
   },
 ])
+
+export const scrollTop = style({
```
chanmioh, line 229: In here, a lot of your styles are overriding each other and you can remove some, such as your `flexDirection` and `flexWrap` since the `flexFlow` overrides this.

website/src/pages/documents/document.page.tsx
chanmioh, line 86: Agree with Naomi, also consider changing this `div` to a `button`, because it has the function of a button!

chanmioh (changes requested):

loafofpiecrust:
A general note: in the future, I wouldn't lump these two changes together into one PR.

website/src/pages/documents/document.css.ts
```diff
@@ -224,3 +224,22 @@ export const mobileWordPanel = style([
...
+  flexDirection: "row",
+  flexWrap: "wrap",
+  flexFlow: "column nowrap",
+  justifyContent: "flex-end",
+  alignContent: "center",
+  alignItems: "center",
+  alignSelf: "flex-end",
+  position: "sticky",
+  zIndex: 1,
+  top: `calc(100vh - 105px)`,
```
loafofpiecrust, line 239: Is there a difference between this and using `bottom: 105`?

website/src/pages/documents/document.css.ts
```diff
@@ -224,3 +224,22 @@ export const mobileWordPanel = style([
...
+  right: 0,
+  padding: `7px`,
```
loafofpiecrust, line 241: Use our constant values from the `style/constants` file like `hspace` and `vspace` for padding.

website/src/pages/documents/document.page.tsx
```diff
@@ -82,6 +83,11 @@ const TabSet = ({ doc }: { doc: Document }) => {
           </Tab>
         </TabList>
       </div>
+      <Button id="scroll-top" className={css.scrollTop} onClick={() => window.scrollTo({top: 0, behavior: 'smooth'})}>
+        <div><a><RiArrowUpCircleFill size={45}/></a></div>
+        <div>Top</div>
```
loafofpiecrust, line 88: What's going on here? A few suggestions: you don't need `<div>` tags everywhere you put content. I don't think you need either here. You also don't need the `<a>` tag which creates a link because this is already inside of a button. Finally, for the text, I'd suggest something that describes the action being performed, like "Scroll to Top". Especially if we only show the text part on desktop, the space difference should be negligible.

website/src/pages/documents/document.page.tsx
```diff
@@ -331,12 +337,20 @@ export const DocumentTitleHeader = (p: {
...
       {p.showDetails ? (
         <Link href={documentDetailsRoute(p.doc.slug!)}>View Details</Link>
       ) : (
         <Link href={documentRoute(p.doc.slug!)}>View Contents</Link>
       )}
+      {!p.doc.audioRecording && (
+        <div id="no-audio-message">
+        {!isMobile ? <strong>No Audio Available</strong> : null}
+        </div>
+      )}
```
loafofpiecrust, line 353: Hmm it looks like this can be simplified by putting the `!isMobile` clause in the first condition, right?

loafofpiecrust (commented):
Here are some more thoughts, @nole2701 reminding you that this PR has several unresolved threads from last month.

website/src/pages/documents/document.css.ts
nole2701: Yes I tried that before, but for some reason it doesn't position right and also stops being sticky as well. I can't exactly figure out why.

GracefulLemming (approved):

PR #236, opened 2022-11-21
Split data validation into separate binary

loafofpiecrust:
Add new CI step for data validation to avoid making the SSH connection wait longer than it has to.


chanmioh (approved):
Looks good!

PR #237, opened 2022-11-29
Fix CWKW bugs

chanmioh:
# Changelog
- Fixed bug where certain Wordpress pages would redirect to the CWKW page (routing issue).
- CWKW page is now accessible by `edited-collections/cwkw`.
- Fixed bug where CWKW banner would appear too large on mobile.

### DIFF
--- a/website/src/pages/cwkw/cwkw-layout.css.ts
+++ b/website/src/pages/cwkw/cwkw-layout.css.ts
@@ -72,7 +72,7 @@ export const banner = style([
   paddingY(vspace.medium),
 
   {
-    width: `calc(${hsize.small} / 3)`,
+    width: hsize.xsmall,
     "@media": {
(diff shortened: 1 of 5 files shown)


website/src/pages/cwkw/cwkw-layout.css.ts
```diff
@@ -70,7 +70,9 @@ export const openHeaderContents = style([
 export const banner = style([
   margin(0),
   paddingY(vspace.medium),
+
   {
+    width: `calc(${hsize.small} / 3)`,
```
loafofpiecrust, line 75: I'd use a fixed value added to the `hsize` object as a new field

chanmioh: fixed!

GracefulLemming (approved):

PR #245, opened 2022-12-08
Redirect old documents to corresponding CWKW page and display Wordpress content on CWKW chapters

chanmioh:
# Changelog
- Added resolver to `AnnotatedDoc` called `chapters` that returns a list of collection chapters that may contain this document, allowing redirection to a collection chapter from a document. 
- Chapters with a Wordpress ID will display the Wordpress page accordingly, given the chapter's url slug

Note: I don't think the deploy preview will work for this PR because I made some backend changes that affects the query for chapters, so the TOC will appear empty in this preview.

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -235,6 +235,7 @@ type CollectionChapter {
 	Full path of the chapter
 	"""
 	path: [String!]!
+	slug: String!
 	document: AnnotatedDoc
 }
 
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -69,6 +69,10 @@ type AnnotatedDoc {
 	These words need to be corrected or reviewed further.
 	"""
 	unresolvedForms: [AnnotatedForm!]!
+	"""
+	Collection chapter path of this document
+	"""
+	chapterPath: [String!]
 }
 
 """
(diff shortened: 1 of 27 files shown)


graphql/schema.graphql
```diff
@@ -69,6 +69,10 @@ type AnnotatedDoc {
 	These words need to be corrected or reviewed further.
 	"""
 	unresolvedForms: [AnnotatedForm!]!
+	"""
+	Collection chapter path of this document
+	"""
+	chapterPath: [String!]
```
loafofpiecrust, line 75: A document isn't guaranteed to live in just one collection

loafofpiecrust (commented):
What's the reasoning for this change? I don't think it fits our data model exactly. We could have a list of chapters that the document lives in, perhaps.

graphql/schema.graphql
chanmioh: Ah okay, my reasoning for the change was that people who have old documents bookmarked can be redirected to the document's corresponding chapter in CWKW. Should they be kept to the same document page instead?

loafofpiecrust: And then just return all of the chapters that this document is attached to. Then on the front-end, we only redirect if there's exactly one chapter the document is within.

loafofpiecrust: If there's multiple, we can show a list of links to them. It also wouldn't be that bad to leave the current document pages up for now too. I originally thought we'd have more collection-specific content associated with each document but we don't have _any_ planned so far.

GracefulLemming (approved):

PR #247, opened 2022-12-09
[D-P171] Move document metadata from a separate page to a tab

nole2701:
Moved the "view details" button contents into a new tab and removed old files/codes for the old button.


GracefulLemming (changes requested):
Please delete website/src/pages/documents/details.page.tsx, there should be no need for it anymore.

GracefulLemming (approved):

PR #248, opened 2022-12-12
[DP-178] As an editor, I need to be able to insert audio for a word or phrase into a Wordpress page

chanmioh:
- Modified regex for inserting audio into Wordpress pages, where you append text `audio` or `audiojoin` to bracket notation to indicate whether you're inserting 1. A single audio track of the words specified 2. Audio tracks for each word specified.
     - Example regex below 
     - [wj46:1-5:audiojoin] 
     - [wj46:1-5:audio] 
     - [wj46:1-5]
- Added this query to retrieve the audio track of a document, necessary when returning a single audio track for the specified words.

Sample page: https://wp.dailp.northeastern.edu/wp-admin/post.php?post=2&action=edit

- One audio track for multiple words: 
image

- Separate audio track for each word: 
image


website/src/components/wordpress.tsx
```diff
@@ -125,9 +211,8 @@ const PullWords = (props: {
   const [{ data }] = Dailp.useDocSliceQuery({
     variables: {
       slug: props.slug,
-      // Convert our inclusive 1-indexed range into an exclusive 0-indexed range.
-      start: props.first - 1,
-      end: props.last ?? props.first,
+      start: props.first,
```
chanmioh, line 214: Reason for change: In the backend, the `start` and `end` indices are ultimately 1-indexed, so I changed this because previously it was converting it to `0-index`.

GracefulLemming (approved):

PR #250, opened 2022-12-16
If a chapter contains a Wordpress page, link to WP via the modified URL slug

chanmioh:
Just replaces "_" with "-" when linking to a CWKW chapter with Wordpress page content, since the Wordpress slug uses dashes instead of underscores.

I.e. CWKW_acknowledgements -> CWKW-acknowledgements

For the Wordpress prose to appear, its slug text should match the one specified in the CWKW index except with underscores rather than dashes.


GracefulLemming (approved):
Approving this so we can show the UKB but in the future we should enforce consistency in our slugs on the backend (all using '-').

PR #252, opened 2022-12-20
[DP-183] Word slices should cite and link their associated document 

hazelyn11:
- Added citation below word slices including document title and word numbers. The citation links back to the document in which the words originated.

One word:
image

Multiple words:
image


website/src/graphql/dailp/queries.graphql
```diff
@@ -259,6 +259,8 @@ query DocSlice(
   $morphemeSystem: CherokeeOrthography!
 ) {
   document(slug: $slug) {
+    title
+    url
```
hazelyn11, line 263: I added "url" here thinking it was something else, and need to remove it, but I keep getting errors when trying to do so even though I've rerun dev-generate-types. VSCode keeps telling me to run dev-check, which I do and get no issues, but then at the very end it tells me to run dev-check again. What should I do?

website/src/components/wordpress.tsx
```diff
@@ -226,19 +227,40 @@ const PullWords = (props: {
...
+
+  let documentCitation = docContents.title + ", word " + props.first
+
+  if (props.last) {
+    documentCitation =
+      docContents.title + ", words " + props.first + " – " + props.last
+  }
+
+  let documentLink =
+    "https://dailp.northeastern.edu/documents/" +
```
hazelyn11, line 239: Will this link be changing once CWKW is launched? Is there an alternative to hard coding it?

website/src/graphql/dailp/queries.graphql
GracefulLemming: Are you still having this issue?

GracefulLemming:
Following up on this PR– @hazelyn11 do you have time to look into the comments above? @chanmioh would you have time to review this?

website/src/components/wordpress.tsx
hazelyn11: that GraphQL call would involve using CollectionChapter, right, since that's where the specific slug is stored? I am looking rn

hazelyn11: I've been looking for a bit-- I think this would be different since the url for CWKW is stored as the path in CollectionChapter and isn't necessarily the slug. I'm a bit confused on how to query CollectionChapter from document (?) or how the logic there would work to get that value. would it involve modifying an existing GraphQL query? Also, right now we only have one collection, CWKW, but as we expand, how will the specific collection name itself be stored? I couldn't find a value for that within CollectionChapter

chanmioh: if i'm understanding your question correctly, yes i would avoid hard-coding this and instead utilize one of the defined routes in the file `routes.ts`. i think you should be able to use the `documentWordPath` route in that file but if the formatting is not the same, you can define a new route in that file to call here!

website/src/components/wordpress.tsx
```diff
@@ -3,6 +3,7 @@ import parse, {
   attributesToProps,
   domToReact,
 } from "html-react-parser"
+import { first } from "lodash"
```
chanmioh, line 6: is this import being used anywhere? otherwise it can be removed!

website/src/pages/documents/document.page.tsx
```diff
@@ -165,7 +165,10 @@ export const TabSet = ({ doc }: { doc: Document }) => {
         {docData.sources.length > 0 ? (
           <section className={fullWidth}>
             Original document provided courtesy of{" "}
-            <Link href={docData.sources[0]!.link}>{docData.sources[0]!.name}</Link>.
+            <Link href={docData.sources[0]!.link}>
```
chanmioh, line 168: here instead of using an "!" i think you should be able to check if the docData.sources[0] itself exists (i think you're already doing that by checking the length, but looks like typescript is finicky) since using a "!" can often lead to unexpected behavior

chanmioh (changes requested):

website/src/components/wordpress.tsx
hazelyn11: I looked in the `routes.ts` file and wrote a new `collectionWordPath` route, but I'm still a little confused about what the best way to access the collection and chapter slug from the `wordpress.tsx` file would be

GracefulLemming (approved):

PR #257, opened 2023-01-20
Convert front-end collection slugs to be lowercase with hyphens

loafofpiecrust:
- Collection slugs in the database (in an ltree) use underscores as a separator, and technically allow upper or lower case letters.
- On the website, we want all URL path segments to be lowercase with hyphens
- This change converts back and forth in the GraphQL layer

### DIFF
--- a/graphql/src/query.rs
+++ b/graphql/src/query.rs
@@ -1,6 +1,6 @@
 //! This piece of the project exposes a GraphQL endpoint that allows one to access DAILP data in a federated manner with specific queries.
 
-use dailp::{CollectionChapter, Uuid};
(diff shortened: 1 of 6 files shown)


chanmioh (approved):
This looks good to me! So if I understood correctly, in the database slugs can have underscores and are not consistently capitalized, so when we want to retrieve them using the resolvers, they should be converted to have "-" instead of "_" and be all lowercase?

loafofpiecrust:
Yes, exactly. In this change, I also added a step to normalize and lowercase slugs from the TOC spreadsheet on ingestion. So "CWKW_introduction" gets turned into "cwkw_introduction" entering the DB, then "cwkw-introduction" in the GraphQL resolver.

PR #261, opened 2023-01-26
Upsert SourceAttribution field into database

chanmioh:
After migration, a document's metadata contains the SourceAttribution field, but this field does not get encoded into the database.

- Added a SQL query for upserting the `document_source` table from a document metadata's SourceAttribution field.
- Added the corresponding DB query to call SQL query.


types/src/database_sql.rs
```diff
@@ -580,6 +580,22 @@ impl Database {
...
 
+        // Upsert document sources based on the SourceAttribution in a document's metadata.
+        {
+            let (name, url): (Vec<_>, Vec<_>) = meta
+                .sources
+                .iter()
+                .map(|source| (&*source.name, &*source.link))
+                .multiunzip();
+            query_file!(
+                "queries/upsert_document_sources.sql",
```
loafofpiecrust, line 591: I think we've got a logic mismatch here -- might need schema changes. So a source link for a particular document points to the specific record in the source archive. What we want is a single row in `document_source` for the Yale Beinecke library and then a `document_source_citation` row for every document that has `sources` filled in. Now `document_source_citation` doesn't have a link/url field right now, I think it needs one added. Does that make sense?

loafofpiecrust (changes requested):

types/src/database_sql.rs
chanmioh: I think I kind of understand, but I'm wondering why does the `document_source_citation` need a link/url field? I'm looking at the docs for the current schema and I see `document_source_citation` has `source_id` to link to the corresponding `document_source` which also has the link/url field you're talking about I assume.

image

PR #262, opened 2023-02-03
Add a CWKW landing page and link landing page to the first chapter

chanmioh:
- Adds CWKW landing page Wordpress content to the homepage of CWKW.
- The "Begin reading" text on the landing page should now link to the first chapter of CWKW (the acknowledgements chapter)


loafofpiecrust (approved):
Decent solution for now -- it could also be simpler to just give the first chapter a special URL rather than doing it in the components. Because also this will generate a duplicate page, which isn't good for SEO.

chanmioh:
# Changelog
- Changed Landing page's Wordpress url to "/cwkw" so it does not get generated by the `chapter.page.tsx`

GracefulLemming:
Merging this in since previous. comments were addressed. Further work can be revisited as needed

PR #266, opened 2023-03-20
Update db setup instructions

CharlieMcVicker:
I was doing developer setup today and I think the name of the `dev-migrate` script must have changed at some point. I ran the following and got this error.
```sh
> dev-migrate
bash: dev-migrate: command not found
```

Now, after running `dev-migrate-schema` and `dev-migrate-data`, my local environment is working great, so I updated the README to have these two commands.

I also wanted to test if I could PR with my permissions on GitHub :)


GracefulLemming (approved):
Thank you so much for raising this!

PR #268, opened 2023-04-06
[DP-186; 200] Contributor audio upload; editor audio curation

CharlieMcVicker:
UX changes:
- Edit word pane now has an "EditWordAudio" section
- - Users with the group "Contributors" are given the ability to upload or record audio, as well as see **all audio** associated with a word (contributor/ingested, curated/not curated).
- - Users with the group "Editors" are shown **all audio** with the ability to check/uncheck if the audio should be shown to readers (ie. if the audio should be included in the edited collection)
- Readers may now see multiple audio recordings for a word

Backend changes:
- Created "curateWordAudioMutation" which requires "Editors" group and updates if audio is included in the edited collection
- Created "attachAudioToWord" mutation that ties a bit of user uploaded audio to a specific word


CharlieMcVicker:
I have since read #212 and am convinced it will merge fine :)

GracefulLemming (approved):

PR #269, opened 2023-04-11
Added compatibility with S3 buckets as audio input sources

GracefulLemming:
Extended data migration to use DAILP's media S3 bucket.
Briefly tested on browser (Safari on Mac) using EFN2 - Child of Igakala as test file.

### DIFF
--- a/migration/src/audio.rs
+++ b/migration/src/audio.rs
@@ -1,38 +1,137 @@
+use anyhow::Ok;
 use dailp::{AudioSlice, DocumentAudioId};
-use itertools::Itertools;
 use log::{error, info};
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap},
     time::Duration,
 };
 use tokio::time::sleep;
 
-/// The string ID for a DRS item, usu. prefixed with "neu:"
+/// Represents an identifier from Northeastern's Digital Repository Service (DRS)
+/// 
+/// DRS identifiers contain a 9-digit base-36 number prefixed with "neu:".
+/// An ID can refer to any kind of DSG object, including collections,
+/// object containers, and raw data (such as text, audio, or video).
+/// 
+/// Examples:
+/// 
+/// The DRS ID for a collection, ...
+/// ```
+/// let drs_collection_id = DrsId("neu:ww72bh375")
+/// ```
+///
+///  ... for an object container holding audio and metadata, ...
+/// ```
+/// let drs_audio_container_id = DrsId("neu:2r36v5312")
+/// ```
+/// 
+/// ... for raw audio data, ...
+/// ```
+/// let drs_audio_data_id = DrsId("neu:4f17r7213")
+/// ```
+/// 
+/// ... and for raw text data.
+/// ```
+/// let drs_text_data_id = DrsId("neu:bz613v117")
+/// ```
 #[derive(Serialize, Deserialize, Clone, Debug)]
 struct DrsId(String);
 
+/// Represents key-value pairs used to structure DRS data.
+/// 
+/// Currently, only supports data in the form `"url": "value"`,
+/// which is used for encoding the canonical object and content objects for a DRS entry.
+/// Note that [ComplexDrsObject]s are a tuple struct wrapping [HashMap].
+/// 
+/// Examples:
+/// 
+/// The following two examples assume an HTTP response with the following JSON snippet:
+/// ```text
+/// {
+///     "https://repository.library.northeastern.edu/downloads/neu:4f18fk930?datastream_id=content": "Audio File"
+/// }
+/// ```
+/// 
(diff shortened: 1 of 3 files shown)


migration/src/audio.rs
```diff
@@ -48,8 +144,15 @@ impl DrsRes {
...
+    fn get_url(&self) -> String {
+        self.canonical_object.get_first_object_pid()
+    }
 }
 
+/// Represents annotation info associated with a single word
+///
+///
 #[derive(Serialize, Deserialize, Clone, Debug)]
 struct AudioAnnotationRow {
```
hazelyn11, line 157: is there a reason this is named Row if it represents a single word?

migration/src/audio.rs
```diff
@@ -58,62 +161,100 @@ struct AudioAnnotationRow {
...
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct AudioRes {
-    /// The URL for a DRS audio resource
+    /// The URL associated with an audio resource.
+    ///
+    /// Currently, this URL will be from either the DRS or an S3 bucket.
     audio_url: String,
-    /// The raw text for a file marking words within a recording
+
+    /// Text holding start and end times for words contained within an audio file
```
hazelyn11, line 174: based on the example it looks like the word itself is also included?

migration/src/audio.rs
GracefulLemming: Good catch. Does this look better?

migration/src/audio.rs
```diff
@@ -1,38 +1,134 @@
...
+/// ... and for raw text data.
+/// ```
+/// let drs_text_data_id = DrsId("neu:bz613v117")
+/// ```
 #[derive(Serialize, Deserialize, Clone, Debug)]
 struct DrsId(String);
 
+/// Represents key-value pairs used to structure DRS data.
+///
+/// Currently, only supports data in the form `"url": "value"`,
```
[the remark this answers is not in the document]
GracefulLemming: Yes

migration/src/audio.rs
```diff
@@ -1,38 +1,134 @@
...
+    }
+}
+
+/// Represents info associated with an object from the DRS.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 struct DrsRes {
     pid: DrsId,
-    // breadcrumbs: ComplexDrsObject,
     parent: DrsId,
     thumbnails: Vec<String>,
```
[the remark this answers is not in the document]
GracefulLemming: These are all fields the DRS provides us, so I have it structured this way to make encoding easier. I will add documentation discussing this

migration/src/audio.rs
GracefulLemming: The text for the word is included in this type (represented by AudioAnnotationRow's "word" field), but is ultimately not used in the creation of an AudioSlice. So, the word data basically gets tossed out for now.

migration/src/audio.rs
```diff
@@ -1,38 +1,174 @@
...
+    /// Creates a new DRS resource
+    ///
+    /// # Errors
+    /// Fails if a connection with the DRS can not be established in a set number of tries
+    /// or if the response body cannot be deserialized properly.
     pub async fn new(client: &Client, drs_id: &str) -> Result<Self, anyhow::Error> {
         let drs = "https://repository.library.northeastern.edu/api/v1/files/";
         let mut tries = 0;
+
         Ok(loop {
```
chanmioh, line 172: when would this loop be broken out of? reading the rust documentation for a loop, it looks like this is an infinite loop until a `break` is called

migration/src/audio.rs
```diff
@@ -1,38 +1,174 @@
...
+    /// }
+    /// ```
+    /// and
+    /// ```text
+    /// {
+    ///     "https://repository.library.northeastern.edu/downloads/neu:4f18fk930?datastream_id=content": "Audio File",
+    ///     "https://repository.library.northeastern.edu/downloads/neu:4f18fk948?datastream_id=content": "Master Image"
+    /// }
+    fn get_first_object_pid(&self) -> String {
+        self.0.keys().next().unwrap().clone()
```
chanmioh, line 109: would there be any cases where the HashMap is empty? since `next()` returns an `Option` type, if the HashMap is empty here than `next()` will return `None` which will cause a panic if `unwrap()` is called on a `None`. I think you could maybe use `unwrap_or_default()` so it won't panic on an empty HashMap

chanmioh: i think you could also make the return type here optional, so it will return a `None` if the HashMap is empty

migration/src/audio.rs
GracefulLemming: When the client returns an ok result or tries exceed 7. See lines 180-183.

migration/src/audio.rs
GracefulLemming: I _believe_ we can trust the HashMap is not empty, but I will make the suggested change anyway for safety!

chanmioh (approved):
looking good to me!

PR #272, opened 2023-04-18
Add creative commons support

GracefulLemming:
Assumption:
- All licensed works DAILP will create will 
  - Require author(s) attribution in compliance with CC-BY
  - Allow only non-commercial use, in compliance with CC-BY-NC

Changes:
- Created a component for adding CC-BY-NC licenses to documents, collections, and other materials.
- Created styles for CC-BY-NC licenses
- Added CC-BY-NC license to DAILP footer
- Added CC-BY-NC license to CWKW ToC


website/src/components/attributions/creative-commons.tsx
```diff
@@ -0,0 +1,62 @@
...
+    title: string,
+    authors: Array<{name: string, link?: string}>
+  }
+
+export const CreativeCommonsBy = (props: Props) => {
+    const licenceAttributes = {
+        'xmlns:cc':"http://creativecommons.org/ns#",
+        'xmlns:dct':"http://purl.org/dc/terms/"
+    };
+    const authors = props.authors.length-1;
```
hazelyn11, line 15: my one suggestion would be to maybe name this constant something else like authors_count? it tripped me up a bit at first since the array is also named authors in the props. super small change tho!

GracefulLemming: Incorporated this in a new commit!

website/src/footer.tsx
```diff
@@ -1,15 +1,29 @@
...
 /** University affiliation, related navigation links, and contact info.  */
 const Footer = () => {
   const { buildDate } = usePageContext()
   return (
     <footer className={css.footer}>
       <div className={css.dark}>
+        <span className={css.content} style={{ display: "block" }}>
+            <CreativeCommonsBy 
+              title="The Digital Archive of Indigenous Language Persistence (DAILP)"
+              authors={
```
chanmioh, line 17: Would these authors possibly change at all in the future? I'm wondering if we would want to store this information in our database rather than hardcoding it here so it can be more flexible to change in the future

website/src/components/toc.tsx
```diff
@@ -60,6 +61,31 @@ const CollectionTOC = () => {
           </Fragment>
         ) : null
       )}
+      <div className={css.noticeText}>
```
chanmioh, line 64: I think maybe instead of putting the Creative Commons License here we could put it at the footer of the pages? To be more consistent with the DAILP pages

website/src/components/toc.tsx
```diff
@@ -60,6 +61,31 @@ const CollectionTOC = () => {
...
+      <CreativeCommonsBy
+        title="Cherokees Writing the Keetoowah Way"
```
chanmioh, line 66: If we can have different authors depending on the collection, we could probably add author information into our database

chanmioh (changes requested):

website/src/footer.tsx
GracefulLemming: No we definitely do want it in the DB. I can make those changes in a new issue and PR maybe? I want to at least get this scaffolding online sooner than later

website/src/components/toc.tsx
GracefulLemming: Im ok adding it to the CWKW footer if @LN7-cyber is!

chanmioh (approved):

PR #273, opened 2023-04-27
DP-197 Word pane and paragraph pane does not change when navigating to a new document

hazelyn11:
- Add useEffect hook to reset state

### DIFF
--- a/website/src/pages/documents/document.page.tsx
+++ b/website/src/pages/documents/document.page.tsx
@@ -182,6 +182,10 @@ export const TranslationTab = ({ doc }: { doc: Document }) => {
     null
   )
 
+  useEffect(() => {
+    selectAndShowContent(null)
+  }, [doc.id])
+
   const [dialogOpen, setDialogOpen] = useState(false)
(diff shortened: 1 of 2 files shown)


GracefulLemming (approved):

website/src/pages/documents/document.page.tsx
```diff
@@ -182,6 +182,10 @@ export const TranslationTab = ({ doc }: { doc: Document }) => {
     null
   )
 
+  useEffect(() => {
+    selectAndShowContent(null)
+  }, [doc.id])
+
```
CharlieMcVicker, line 188: Do we have any preference for this over setting a `key` on this component, so that there is a full remount when the document id changes?

CharlieMcVicker (approved):
This looks good, just one idea.

website/src/pages/documents/document.page.tsx
GracefulLemming: Setting a key is probably good practice

PR #279, opened 2023-05-23
Add-word-search-to-cwkw-index

GracefulLemming:
## Changes
- Creates a component for word search
- Adds Word Search on CWKW index
- Removed normalization field from search

## Future work: 
- Allow alphabetical sorting for each column
- allow search to be constrained to the scope of a single edited collection


website/src/components/wordpress.tsx
```diff
@@ -69,33 +70,39 @@ export const WordpressPageContents = ({
 
 const parseOptions: HTMLReactParserOptions = {
   replace(node) {
-    const style = /\[(\w*):([0-9]*)-?([0-9]*)?:?(audio)?(join)?\]/ // [DocName:Start(-OptionalEnd):?(audio?)(join?)]
+    const wordEmbedStyle = /\[(\w*):([0-9]*)-?([0-9]*)?:?(audio)?(join)?\]/ // [DocName:Start(-End?):?(audio?)(join?)]
+    const referenceEmbedStyle = /\[(\w*)\]/ // [search | glossary]
```
hazelyn11, line 74: So through this syntax, we can insert a lexical search anywhere using Wordpress? Am I reading this code correctly?

website/src/components/wordpress.tsx
```diff
@@ -69,33 +70,39 @@ export const WordpressPageContents = ({
...
+            first={parseInt(wordSegments[2]!)}
+            last={wordSegments.length >= 4 ? parseInt(wordSegments[3]!) : undefined}
           />
         )
       }
+
+      if (referenceSegments && referenceSegments[1] === "search") {
+        return (<LexicalSearch/>)
+      }
     }
```
hazelyn11, line 107: Am I just missing it or is there no case for "glossary" yet?

hazelyn11: In the comment by the syntax it looks like both "search" and "glossary" will eventually be supported

hazelyn11:
I added the search syntax to the feature preview page and it works!! My one question is if this PR will also support glossaries added through WP, because it looked like the case for that was missing within wordpress.tsx if so, although there was a comment that suggested it was supported

website/src/components/wordpress.tsx
GracefulLemming: There is not, that will be in another PR today

hazelyn11 (approved):

PR #280, opened 2023-05-25
[DEV ONLY] Cognito identity JWT auth for GQL

CharlieMcVicker:
# Problem to solve
We need to send Cognito credentials to our backend to properly restrict user access based on Cognito role.

# Solution
- Send these credentials from the client via an `urql` exchange, called `authExchange`, via Cognito identity JWTs
- Validate JWTs manually on the server-side, because there are not stable AWS rust bindings

# Follow up (in a future PR)
- Figure out what needs to happen in production, with the `lambda_http` server.


graphql/src/cognito.rs
```diff
@@ -0,0 +1,43 @@
+use crate::query::UserInfo;
+use jsonwebtoken::{jwk::JwkSet, Validation};
+
+/// Load the set of keys that can be used to validate Cognito identity tokens
+fn get_jwk_set() -> Result<JwkSet, serde_json::Error> {
+    // this file is pulled from:
+    // https://cognito-idp.<Region>.amazonaws.com/<userPoolId>/.well-known/jwks.json
+    serde_json::from_str::<JwkSet>(include_str!("dailp_user_pool_jwks.json"))
```
GracefulLemming, line 8: Should the contents of the JWK Set be in the repo? Id imagine GETting the url listed above would be a safer (and more maintainable) solution.

website/src/renderer/_default.page.client.tsx
```diff
@@ -1,13 +1,23 @@
...
 import ReactDOM from "react-dom"
+import { Client } from "urql"
+import { getCredentials } from "src/auth"
 import { customClient, sharedSsr } from "src/graphql"
 import { PageContext, PageShell, rootElementId } from "./PageShell"
 
-const client = customClient(false, [sharedSsr])
+async function getClient() {
+  const token = await getCredentials()
+  console.log("[charlie]", { token })
```
GracefulLemming, line 10: Maybe delete this line if it was just to help in your debugging

GracefulLemming (approved):

GracefulLemming (approved):

PR #296, opened 2023-07-24
[DP-212] Scroll to top should be hidden on pages that are too short to scroll

nole2701:
Made scroll to top button pop in and out when window.scrollY crosses between 2000px.

I decided on 700px after scrolling through some pages on mobile mode and it seems like 2000px is a swipe or two.


website/src/pages/documents/document.css.ts
```diff
@@ -234,6 +234,48 @@ export const mobileWordPanel = style([
...
+//   padding: vspace.quarter,
+//   color: colors.secondaryContrast,
+//   outlineColor: colors.secondaryContrast,
+//   backgroundColor: colors.secondary,
+//   visibility: "visible",
+//   opacity: 1,
+//   transition: "top 0.5s",
+//   }
+// )
+
```
GracefulLemming, line 278: if these commented lines are not to be used, please delete them!

GracefulLemming (changes requested):

PR #320, opened 2023-09-29
[GQL] Comments

CharlieMcVicker:
# Completed user stories:

- [x] As a DAILP user, I can leave a text comment on a word or paragraph I am viewing, so that I can share my knowledge in whatever form it is

- [x] As a user, I can read other users comments about a word or paragraph I am viewing

- [x] As a user, I can delete my own comments, so I can fix typos and inaccuracies and also have control over my own contributions

# Todo

- [x] We could document the new SQL enum types and their corresponding rust types a bit better 
- [x] As a user, I want the option to add a tag or flair to my comment so people know how I am hoping to contribute (story, correction, etc.)

There is support for this in the Create/Read operations, but the open question is: **what are the tags we will support?**


types/src/database_sql.rs
```diff
@@ -46,6 +47,75 @@ impl Database {
...
+        Ok(query_file_as!(
+            BasicComment,
+            "queries/comments_by_parent.sql",
+            parent_id,
+            parent_type.clone() as CommentParentType
+        )
+        .fetch_all(&self.client)
+        .await?
+        .into_iter()
+        .map(|c| c.into())
```
CharlieMcVicker, line 75: There may be syntactic sugar for this (Like `Into::into` or something) but I couldn't get it to work.

types/src/lib.rs
```diff
@@ -38,6 +38,7 @@ mod slugs;
 mod tag;
 mod translation;
 pub mod user;
+pub mod comment;
```
CharlieMcVicker, line 41: I've noticed my imports from this module have to be qualified an extra time when they are imported (`dailp::comment::whatever`) is that because I used `pub mod ...`? Should I get rid of `pub`?

graphql/schema.graphql
```diff
@@ -404,6 +457,27 @@ input DateInput {
 	year: Int!
 }
 
+type DateTime {
```
GracefulLemming, line 460: Is there a reason you aren't using/modifying the existing date datatype?

https://github.com/NEU-DSG/dailp-encoding/blob/f5ca41369dbf251e714a48d9ea4fbbdf240d7cd4/graphql/schema.graphql#L382

graphql/schema.graphql
```diff
@@ -455,16 +529,30 @@ type DocumentPage {
 	paragraphs: [DocumentParagraph!]!
 }
 
+"""
+One paragraph within a [`DocumentPage`]
+"""
 type DocumentParagraph {
 	"""
-	Source text of the paragraph broken down into words
+	Database ID
```
GracefulLemming, line 537: Not sure how other people would feel abt this, but at least for me it took a minute to comprehend what this meant. I initially interpreted it at `an id representing a database` instead of the intended `an id for referencing this paragraph within our database`

GracefulLemming (changes requested):

types/src/lib.rs
GracefulLemming: Yeah, we keep everything package private. It also helps prevent crossing wires in other rust code which is nice

types/src/database_sql.rs
GracefulLemming: I think this is perfectly fine tbh, it feels clear on first glance so I dont think any syntactic sugar would be better

graphql/schema.graphql
CharlieMcVicker: Date wraps a "Date," the new DateTime type wraps a "DateTime"

graphql/schema.graphql
CharlieMcVicker: Updated.

GracefulLemming (approved):

PR #351, opened 2023-10-27
Add duration bounds checking to word-for-word audio migration

CharlieMcVicker:
# Goal
The purpose of this PR is to catch mismatched annotations and audio files at migration time.

## Checks
- [x] Annotations do not reference audio outside of the bounds of the audio file's duration
   - Sadly this involved downloading the audio to tmp files and doing the duration check on disk, which is not ideal for the long run. I'd like to find a rust crate that will do the metadata read over the network, but I couldn't find one quickly and moved on. I consider this a quick fix and diagnostic while we are still getting to the bottom of things.
- [ ] Some kind of check which can handle when a **smaller annotation** file is used. Eg. more than 10 seconds between the last annotation `end_time` and the duration of the file. For false positives, we would probably want to trim the end of the audio file anyway. Removing silence at the end of a track wouldn't affect alignment.
- [ ] There are the same number of annotations and words in a document

## Counting words / annotations and ensuring they match
I couldn't find a clean place to add a sanity check on `num_words == num_annotations` but here is the code for it. I originally wrote this inside `AnnotatedLine::many_from_semantic` [source](https://github.com/NEU-DSG/dailp-encoding/blob/cm/audio-bounds-checking/migration/src/spreadsheets.rs#L577-L689) but since that goes line by line, this would typically fail, since there are fewer words in a line than in the whole document.

```rust

                // Our code will error if there are fewer annotations than
                // words, but continue silently if there are leftover
                // annotations.
                if let (Some(num_annotations), Ok(num_words)) = (
                    meta.audio_recording
                        .as_ref()
                        .and_then(|audio| audio.annotations.as_ref())
                        .map(|annotations| annotations.len()),
                    words.as_ref().map(|w| w.len()),
                ) {
                    anyhow::ensure!(
                        num_annotations == num_words,
                        "If we have word-for-word annotations, we must have one for each word."
                    )
                }
```

### DIFF
--- a/migration/src/audio.rs
+++ b/migration/src/audio.rs
@@ -4,7 +4,7 @@ use log::{error, info};
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::{collections::HashMap, time::Duration};
+use std::{collections::HashMap, io::Cursor, time::Duration};
 use tokio::time::sleep;
 
 /// Represents an identifier from Northeastern's Digital Repository Service (DRS)
@@ -214,6 +214,9 @@ pub struct AudioRes {
     /// Currently, this URL will be from either the DRS or an S3 bucket.
     audio_url: String,
 
+    /// The duration of the associated audio file, for bounds (and sanity) checking.
+    audio_duration: Duration,
+
     /// Text holding start and end times for words contained within an audio file
     ///
     /// Example Contents:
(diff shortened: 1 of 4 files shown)


LN7-cyber:
Thank you, Charlie. Please keep at it.

GracefulLemming:
If we dont want to download each audio file, two things come to mind:
1. Track audio length on a custom metadata field in S3
2. implement a flag representing whether the length of an audio file has already been verified since last update, allowing us to skip downloading some audio

Both of these come with their own overhead in ways im not certain are direct improvements on this process though

migration/src/audio.rs
```diff
@@ -282,21 +296,39 @@ impl AudioRes {
...
-        AudioSlice {
+    pub fn into_document_audio(self) -> Result<AudioSlice, anyhow::Error> {
+        let resource_url = self.audio_url.clone();
+        let audio_duration = self.audio_duration.clone();
+        let annotations = self.into_audio_slices();
+
+        if let Some(slices) = &annotations {
+            let last_slice = slices
+                .last()
+                .ok_or_else(|| anyhow::format_err!("Expected there to be at least one slice"))?;
```
GracefulLemming, line 307: Ive noticed that sometimes our logging appears in unpredictable orders due to use of async, so I think adding some info identifying the problematic document, annotations, and/or audio file as well as the particular word index and/or annotation time/index would be helpful in resolving errors thrown here

PR #352, opened 2023-10-30
Dashboard bookmarks tab.

nole2701:
- Added /dashboard page to house the dashboard.
- Added migration for bookmarks in user, and last_edited in document
- Added queries to add/remove bookmarks, and edited document query to include last_edited
- Finished functionality for users to add/remove bookmarks on document pages.
- Mostly completed the bookmarks tab for the dashboard. Missing routing, number of contributions, and last_edited since they aren't implemented yet.

Trying to merge this before starting to add code for routing because all of this wants to use the routing changes too.


graphql/src/query.rs
```diff
@@ -292,107 +291,42 @@ impl Query {
...
+        // let user = context
+        //     .data_opt::<UserInfo>()
+        //     .ok_or_else(|| anyhow::format_err!("User is not signed in"))?;
         Ok(context
             .data::<DataLoader<Database>>()?
             .loader()
-            .update_contributor_attribution(contribution)
+            .get_bookmarks(
+                Uuid::parse_str("910a22ac-21cd-4398-b656-1039931de29f")
+                    .unwrap_or_else(|_| panic!("Failed to parse UUID")),
```
CharlieMcVicker, line 305: Hardcoded UUID? I changed this to get the logged-in user's ID on my branch. I'm guessing you were just testing something, but I thought I'd flag it to make sure this gets updated.

graphql/src/query.rs
```diff
@@ -292,107 +291,42 @@ impl Query {
...
-    #[graphql(
-        guard = "GroupGuard::new(UserGroup::Editors).or(GroupGuard::new(UserGroup::Contributors))"
-    )]
-    async fn delete_contributor_attribution(
-        &self,
-        context: &Context<'_>,
-        contribution: DeleteContributorAttribution,
-    ) -> FieldResult<Uuid> {
+    /// Gets the short name of a document given its id
+    async fn get_doc_short_name(&self, context: &Context<'_>, doc_id: Uuid) -> FieldResult<String> {
```
CharlieMcVicker, line 311: Like I said on the schema file, this should be a field resolver on the `AnnotatedDoc` type, not a top-level query.

graphql/src/query.rs
```diff
@@ -450,6 +384,22 @@ impl Mutation {
             .await?)
     }
 
+    /// Adds a bookmark to the user's list of bookmarks
```
CharlieMcVicker, line 387: Does this function also "un-bookmark" stuff? If so, that should probably go in the doc string and the name should maybe change.

result
CharlieMcVicker, line 1: This file should not be in this PR

tee
CharlieMcVicker, line 1: This file should not be in this PR

cargo-run-gql-schema.txt
CharlieMcVicker, line 1: This file should not be in this PR

types/migrations/20230925072747_add_bookmarked_documents.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+alter table dailp_user
+  add column bookmarked_documents uuid[] not null default '{}';
```
CharlieMcVicker, line 3: Like we mentioned in standup, we should use a join table here, something like what I wrote below:
```sql
create table user_bookmarked_document (
    document_id uuid not null references document (id) on delete cascade,
    user_id uuid not null references dailp_user (id) on delete cascade,
    bookmarked_on date not null DEFAULT now(),
    primary key (document_id, user_id) -- ensures that there is only one row per user/document pair 
);
```

types/queries/add_bookmark.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE dailp_user
+SET bookmarked_documents = CASE
+    WHEN $3 = true AND NOT $2 = ANY(bookmarked_documents) THEN array_append(bookmarked_documents, $2)
+    WHEN $3 = false THEN array_remove(bookmarked_documents, $2)
+    ELSE bookmarked_documents
+  END
+WHERE id = $1;
```
CharlieMcVicker, line 7: This will probably need to be more like

```sql
insert into user_bookmarked_audio
  (user_id, document_id)
  values ($1, $2)
  on conflict do nothing;
```
Then you would write a separate thing to "delete from" etc. where "user_id,document_id" match.

types/queries/get_bookmarks.sql
```diff
@@ -0,0 +1,3 @@
+SELECT bookmarked_documents
+FROM dailp_user
+WHERE id = $1;
```
CharlieMcVicker, line 3:
```sql
select ... -- copy and paste fields from some other query that fetches documents
from user_bookmarked_document
join document on document.id=user_bookmarked_document.document_id
where user_bookmarked_document.user_id=$1
```

types/src/document.rs
```diff
@@ -71,6 +70,11 @@ impl AnnotatedDoc {
         &self.meta.date
     }
 
+    /// Date that the last edit was made to this document
+    async fn last_edited(&self) -> &Option<Date> {
+        &self.meta.last_edited
+    }
+
```
CharlieMcVicker, line 77: this is where a function to get the document short name should live

types/src/user.rs
```diff
@@ -20,3 +21,12 @@ pub struct User {
...
 }
+
+/// Adds a bookmarked document to the user's list of bookmarks
+#[derive(async_graphql::InputObject)]
+pub struct AddBookmark {
+    /// ID of the document to add bookmark
+    pub document_id: Uuid,
+    /// Whether the document is already bookmarked
+    pub bookmark_bool: bool,
+}
```
CharlieMcVicker, line 32: I think `AddBookmark` is not a great name for something that adds or removes a bookmark. `UpdateBookmark` maybe?

CharlieMcVicker (changes requested):
Okay so this is a bunch of stuff. The point of this is to raise issues about long-term maintainability and design. For the most part this is really good but some common things I noticed:
1. Not taking advantage of GraphQL.
  - The top-level query to get information about a document is no good. Ask for the document, ask for the short name on that document. To implement that, you just write a function inside an `impl` block for the `AnnotatedDoc` and do your database query in there.
  - Also, mutations should **never** return plain old UUIDs. The front-end client (urql) will not know what entity has been updated and will not refresh efficiently. Return a full object appropriate for the mutation, like a document.
  - Similarly, returning a `UUID` when retrieving the bookmarks feels off. GraphQL is about providing your data as a graph. UUIDs can't take someone anywhere unless they make another query. You should instead return the object the UUID is tied to, so the person querying can specify the fields they want from that object in a single query.
  - Something I expected to see but wasn't here was a "bookmarked" / "bookmarkedOn" field on the document itself. Looking this up in the list of the user's documents every time doesn't feel like a good solution for checking if the document is already bookmarked on the front-end. This should be done with a field resolver (function in an impl block) that runs an SQL query against the join table to find a row with the given user and document IDs.
2. Not taking advantage of SQL. A join table is really a must here. I know we already talked about it today so I won't really say more, but I think that's gotta happen before we merge this.

graphql/schema.graphql
```diff
@@ -984,6 +820,14 @@ type Query {
 	Basic information about the currently authenticated user, if any.
 	"""
 	userInfo: UserInfo!
+	"""
+	Gets all bookmarks for the currently authenticated user.
+	"""
+	getBookmarks: [UUID!]
```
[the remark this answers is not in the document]
nole2701: With 2, I think leaving it as returning the document id list could make it reusable if need be. Unless there's a large disadvantage to doing so that I don't know about.

GracefulLemming:
Replaced by
- https://github.com/NEU-DSG/dailp-encoding/pull/358
- https://github.com/NEU-DSG/dailp-encoding/pull/356

PR #358, opened 2023-11-10
added join table for bookmarks into the back end

nole2701:
Added a join table to store bookmarks for documents. It has document_id and user_id as primary keys, and another field for the date of the bookmark.

- Added migration file to create the join table
- Added 2 SQL files to add and remove bookmarks
- Linked those files through database_sql.rs, query.rs, to queries.graphql ready to use in the front-end

For testing I needed to hard code the user_id in query.rs but the way user_id is dynamically obtained is the same as other queries in query.rs so it will work once the website is launched.


types/queries/add_bookmark.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO user_bookmarked_document (document_id, user_id, bookmarked_on)
+VALUES ($1, $2, NOW());
```
CharlieMcVicker, line 2: I think you might want "on conflict do nothing" at the end here. This would make sure this doesn't cause an error if a document is already bookmarked.

Also if you want, you could remove `bookmarked_on` and `NOW()` since you set that as the default right?

types/src/database_sql.rs
```diff
@@ -615,6 +615,20 @@ impl Database {
 
     // pub async fn maybe_undefined_to_vec() -> Vec<Option<String>> {}
 
+    pub async fn add_bookmark(&self, document_id: Uuid, user_id: Uuid) -> Result<String> {
+        query_file!("queries/add_bookmark.sql", document_id, user_id)
+            .execute(&self.client)
+            .await?;
+        Ok(format!("document: {}, user: {}", document_id, user_id))
```
CharlieMcVicker, line 622: We should return the `AnnotatedDoc` right?

types/src/database_sql.rs
```diff
@@ -615,6 +615,20 @@ impl Database {
...
+    }
+
+    pub async fn remove_bookmark(&self, document_id: Uuid, user_id: Uuid) -> Result<String> {
+        query_file!("queries/remove_bookmark.sql", document_id, user_id)
+            .execute(&self.client)
+            .await?;
+        Ok(format!("document: {}, user: {}", document_id, user_id))
```
CharlieMcVicker, line 629: Same here re: return types

CharlieMcVicker (changes requested):
This looks great! A couple things still though:

1. The return types on these mutations are a little weird... I imagine that bookmarking a document should return that document, right? That way the frontend can get an updated version of the document that is now bookmarked when it runs that query.

2. Related to that I think that the `bookmarked_on` column should get pulled from the database somewhere? Maybe there should be a resolver on `AnnotatedDoc` which looks up this column? This way your frontend code will be able to know if a document has already been bookmarked. Ie. something like this when you query a document, and you know `null` implies the document is unbookmarked, and a date implies it has been bookmarked? (If you want sample code for making a resolver, the contributors field on `AnnotatedDoc` is a [great example](https://github.com/NEU-DSG/dailp-encoding/blob/bf65689e925c9ec5055ff91ca9685b29456768ab/types/src/document.rs#L97-L108))
```gql
{
  id
  title
  // etc
  bookmarkedOn
}
```

3. Perhaps this PR should also include your top level query for bookmarked documents, ie.
```gql
type Query {
  // ...
  """
  All documents the currently signed in user has bookmarked.
  """
  bookmarkedDocuments: [AnnotatedDoc!]!
  // ...
}
```

types/src/database_sql.rs
nole2701: This matches with what we're doing in other places, such as update_contributor_attribution or update_word. We're returning ids.

types/queries/get_bookmarked_on.sql
```diff
@@ -0,0 +1,3 @@
+SELECT bookmarked_on, document_id, user_id
```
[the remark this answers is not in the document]
CharlieMcVicker: I wonder if `get_document_bookmarked_on` is a good name? I know that's even longer, but I think that expresses the intent of this query. If we have a document and want to know when a user bookmarked it, here is how we find out.

types/queries/all_documents.sql
```diff
@@ -20,4 +21,7 @@ from document
   left join
     contributor_attribution on contributor_attribution.document_id = document.id
   left join contributor on contributor.id = contributor_attribution.contributor_id
-group by document.id
+  left join user_bookmarked_document as ubd on ubd.document_id = document.id
```
CharlieMcVicker, line 24: I think this should be removed now that we are fetching the `bookmarked_on` field lazily, right??

FWIW, if you were going to fetch `bookmarked_on` for many documents, you'll have to also pass in the `id` of the signed-in user and add a condition to the where clause. Ie. something like `AND ubd.user_id = $whatever`. Otherwise this query might start doing weird things once multiple users bookmark the same document.

types/queries/get_bookmarked_on.sql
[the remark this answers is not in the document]
nole2701: I'll change it to `get_document_bookmarked_on` since it is indeed the intent of the loader as a whole. The reason the query needs to return all columns in the table is so it can properly insert it into the HashMap type required as a return for Loaders.

graphql/src/query.rs
```diff
@@ -474,31 +475,47 @@ impl Mutation {
...
+            .data::<DataLoader<Database>>()?
+            .load_one(dailp::DocumentId(document_id))
+            .await?;
+        context
             .data::<DataLoader<Database>>()?
             .loader()
             .add_bookmark(document_id, user.id)
-            .await?)
+            .await?;
+        Ok(document)
```
CharlieMcVicker, line 496: I think we probably want to load the document after we change the data, right? I know in this case it won't matter because we fetch `bookmarked_on` lazily, but it just feels backwards, and it isn't a pattern we should replicate elsewhere.

That would just mean flipping the data loader call and the `add_bookmark` call:

```
        context
            .data::<DataLoader<Database>>()?
            .loader()
            .add_bookmark(document_id, user.id)
            .await?;
        Ok(context
            .data::<DataLoader<Database>>()?
            .load_one(dailp::DocumentId(document_id))
            .await?)
```

graphql/src/query.rs
```diff
@@ -474,31 +475,47 @@ impl Mutation {
...
+            .data::<DataLoader<Database>>()?
+            .load_one(dailp::DocumentId(document_id))
+            .await?;
+        context
             .data::<DataLoader<Database>>()?
             .loader()
             .remove_bookmark(document_id, user.id)
-            .await?)
+            .await?;
+        Ok(document)
```
CharlieMcVicker, line 518: Same thing re: order of doing the mutation (actually changing data) and fetching from the database.

types/src/database_sql.rs
```diff
@@ -1910,14 +1910,18 @@ impl Loader<BookmarkedOn> for Database {
...
         let document_id: Vec<_> = keys.iter().map(|k| k.0).collect();
         let user_id: Vec<_> = keys.iter().map(|k| k.1).collect();
-        let items = query_file!("queries/get_bookmarked_on.sql", document_id[0], user_id[0])
-            .fetch_all(&self.client)
-            .await?;
+        let items = query_file!(
+            "queries/get_document_bookmarked_on.sql",
+            document_id[0],
+            user_id[0]
+        )
```
CharlieMcVicker, line 1917: Something feels really off here. If this accepts a vec of keys we shouldn't implement to fetch just the first key. I'm not sure we really need a dataloader for this either for what it's worth. A data loader is normally for batching big operations. This is a small operation. If your query can only handle looking up one document at a time, you aren't writing a dataloader, it's just a normal query!

Did you model this off another dataloader that does this `[0]` stuff?

That is definitely a bug waiting to happen. All I need to do is call this with more than one document and it'll just fetch the first one and say the second doesn't exist.

website/src/graphql/dailp/queries.graphql
```diff
@@ -415,11 +415,67 @@ mutation CurateWordAudio($input: CurateWordAudioInput!) {
 }
 
 mutation AddBookmark($documentId: UUID!) {
-  addBookmark(documentId: $documentId)
+  addBookmark(documentId: $documentId) {
```
CharlieMcVicker, line 418: When you actually do your frontend PR, make sure to create a fragment on `AnnotatedDoc` with these fields. You should never have to copy+paste a list of fields to resolve. Let me know if you need help doing that.

graphql/schema.graphql
```diff
@@ -817,11 +817,11 @@ type Mutation {
 	"""
 	Adds a bookmark to the user's list of bookmarks.
 	"""
-	addBookmark(documentId: UUID!): String!
+	addBookmark(documentId: UUID!): AnnotatedDoc
```
CharlieMcVicker, line 820: Why is the return type optional? What does it mean if this returns null?

If returning null is something that can really happen and means something, that should be a part of the docstring. If returning null means you gave a bogus document ID, I feel like that should be an error and not returning null. In that case I think you should make this return `AnnotatedDoc!` / remove the `Option<>` in your rust code, and handle the branch where we can't find the document.

CharlieMcVicker (changes requested):
Hey! This is _soooooo_ close. I think the biggest thing here is the dataloader usage. A dataloader is a utility for writing _batched_ loaders, ie. "get me a bunch of documents based on their ids." I don't think this is your use case, and just fetching the first thing in the batch is no good. If someone ever used this dataloader to call `load_many` it wouldn't do the right thing - and that's how bugs happen!

If you want a function that takes in a document ID and user ID and returns when the user bookmarked that document, add a `get_document_bookmarked_at` function to the `Database` implementation where `add_bookmark` is, and take in those two IDs as normal function parameters.

Hopefully, this should be a pretty quick fix! Please message me on teams if any of this is confusing. We're almost there!

I also left one tip for when you do the frontend stuff based on the filler code you left in queries.graphql, just for when you get to that.

types/src/database_sql.rs
nole2701: There is only going to be one element in the vector since the query in the front-end only ever sends in one of each. I'm pretty sure this is currently only used in the AnnotatedDoc query which has no manual input, so it would never have more than one thing in there. But I can see that it could be a problem later if someone manually puts in more than 1. I kind of remember trying to do this without sending in a vector but it gave an error saying it had to be in a vector so I believe that's why I made it like this. I can look at this again tonight when I return.

graphql/schema.graphql
nole2701: There is the option to return null because the document loader used in this returns an Option<AnnotatedDoc>

CharlieMcVicker: The `ok_or` function might help. That turns an Option into a Result. That could quickly turn `None` into an error for us I think

types/src/database_sql.rs
CharlieMcVicker: Yeah the `DataLoader` stuff will always want it to be a vector. Just make a simple function on `Database` in that big `impl` block with all the others. `DataLoaders` can be great, but they solve a very specific problem.

GracefulLemming: @nole2701 are you planning on leaving this as a data loader or will you be making this a function on Database (lines 35-1468)?

GracefulLemming: Actually, I just noticed [this related discussion](https://github.com/NEU-DSG/dailp-encoding/pull/358#discussion_r1407188581), so you really should make this a function on `Database` rather than a data loader imo

GracefulLemming (commented):
As far as I can see it looks like most of the comments that needed addressed have been handled. We need to take a look as a group about where we are returning strings in `database_sql.rs`. Once all of Charlie's review is complete I think we should be ok to merge.

GracefulLemming (approved):

types/src/database_sql.rs
```diff
@@ -637,6 +637,25 @@ impl Database {
...
+            document_id,
+            user_id
+        )
+        .fetch_optional(&self.client)
+        .await?
+        {
+            Ok(Some(date::Date(bookmark.bookmarked_on)))
+        } else {
+            Ok(None)
+        }
```
CharlieMcVicker, line 656: I think you could clean this up with `query_file_scalar!` (this tells sqlx you are only fetching one column, and so it doesn't make a row struct for you) or by "mapping" over your option instead of doing `if let`. Also, I don't think this is worth holding you up over, this is just so you know about this going forward.

Here is the reference for query scalar (query file scalar just takes a file path instead) https://docs.rs/sqlx/latest/sqlx/macro.query_scalar.html 

 I think `query_file_scalar!` would be cleaner in this case but I think it's worth showing what the solution with map would look like if you're ever in this situation again. Here it is with map:

```suggestion
        Ok(
        query_file!(
            "queries/get_document_bookmarked_on.sql",
            document_id,
            user_id
        )
        .fetch_optional(&self.client)
        .await?
        .map(|bookmark| date::Date(bookmark.bookmarked_on))
        )
```

Here are the docs for "map" on the Option type. https://doc.rust-lang.org/std/option/enum.Option.html#method.map

graphql/src/query.rs
```diff
@@ -480,20 +480,20 @@ impl Mutation {
...
             .data::<DataLoader<Database>>()?
             .loader()
             .add_bookmark(document_id, user.id)
             .await?;
-        Ok(document)
+        Ok(context
+            .data::<DataLoader<Database>>()?
+            .load_one(dailp::DocumentId(document_id))
+            .await?
+            .ok_or_else(|| anyhow::format_err!("Failed to load document"))?)
```
CharlieMcVicker, line 496: Awesome

graphql/schema.graphql
```diff
@@ -817,11 +817,11 @@ type Mutation {
 	"""
 	Adds a bookmark to the user's list of bookmarks.
 	"""
-	addBookmark(documentId: UUID!): AnnotatedDoc
+	addBookmark(documentId: UUID!): AnnotatedDoc!
 	"""
 	Removes a bookmark from a user's list of bookmarks
 	"""
-	removeBookmark(documentId: UUID!): AnnotatedDoc
+	removeBookmark(documentId: UUID!): AnnotatedDoc!
```
CharlieMcVicker, line 824: Yay! I love non-nullable types.

CharlieMcVicker (approved):
Woohoo! I left one thing to read over but it's just about cleaning things up if you'd like to.

PR #359, opened 2023-11-14
Add-user-signup-page

GracefulLemming:
This PR accomplishes two major things:
1. Users can create accounts via the website
2. Users must verify their email upon account creation

There are a number of nice-to-haves that I did not take the time to implement for this ticket:
- [ ] Verified users are automatically signed in
- [ ] Users must reenter password to create an account
- [ ] Users can see the email they created an account with when on the verification screen
- [ ] Users can quickly navigate back to the "create account" screen if they made a typo in their email

All reviewers on the dev team should (a) check out this branch and run the PR locally, stepping through the new features, _and_ (b) look through the code and ask themselves "If I had to work with these files in two weeks, could I understand the code written enough to get started?" Your comments on this PR should reflect your thoughts as you go through each file and through your local website instance. Some things you _may_ find include logic mistakes, unclear naming, inconsistent design patterns, or ways to improve the user experience. 

@LN7-cyber: Could you click "Log In" on the linked site preview and try to create an account, then leave a comment discussing your impressions of the experience?

@CharlieMcVicker: Could you specifically double check my auth flow work in `auth.tsx`?


GracefulLemming:
> ### ✅ Deploy Preview for _dailp_ ready!
> Name	Link
> 🔨 Latest commit	
> 🔍 Latest deploy log	https://app.netlify.com/sites/dailp/deploys/6557be4c3209380f2e29afc5
> 😎 Deploy Preview	https://deploy-preview-359--dailp.netlify.app
> 📱 Preview on mobile	
>  Toggle QR Code...
> _To edit notification comments on pull requests, go to your [Netlify site configuration](https://app.netlify.com/sites/dailp/configuration/deploys#deploy-notifications)._

@LN7-cyber Try testing the signup using this link please

hazelyn11:
I am beginning my review now and got this message when I tried to create an account: 
image
I will look through the code and leave comments on the code now!

GracefulLemming:
> I am beginning my review now and got this message when I tried to create an account:
> image
> I will look through the code and leave comments on the code now!

This is an ongoing issue with AWS resources affecting all of our branches, im working on it rn

migration/src/audio.rs
```diff
@@ -229,35 +229,42 @@ impl AudioRes {
     /// Creates a new Audio Resource from an audio key and an annotaion key.
     ///
     /// Keys should either be a resource string from the DRS or DAILP audio S3 Bucket.
-    pub async fn new(audio_ref_key: &str, annotation_ref_key: Option<&String>) -> Result<Self, anyhow::Error> {
+    pub async fn new(
+        audio_ref_key: &str,
+        annotation_ref_key: Option<&String>,
+    ) -> Result<Self, anyhow::Error> {
```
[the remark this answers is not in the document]
GracefulLemming: This particular file is just formatting yes. Youre also right that User Auth Layout provides common components for signin, reset, verification, and signup.

`website/src/auth.tsx` is responsible for handling AWS signin actions, which includes redirecting users to a page based on their signin status. ex. a user is redirected to verification after successfully submitting the signup form

website/src/pages/auth/user-auth-layout.tsx
```diff
@@ -165,4 +110,52 @@ export const FormFields = ({
...
+      )}
+    </>
+  )
+}
+
+export const SignupLink = () => {
+  return (
+    <>
+      <label>
+        Dont have an account? <Link href="/auth/signup">Sign Up</Link>
```
CharlieMcVicker, line 147: I'm not sure `label` is semantically correct here. I think `label` is just for form inputs.

website/src/auth.tsx
```diff
@@ -64,6 +68,54 @@ export const UserProvider = (props: { children: any }) => {
...
+        let cognitoUser = result?.user
+        if (!cognitoUser) {
+          console.log("failed to create user and activate session")
+          return
+        }
+        setUser(cognitoUser)
+        await navigate("/auth/confirm-signup")
+      }
+    )
+  }
```
CharlieMcVicker, line 95: I think you should rename `username` to `email` here, and comment where you use it as a username and say "we are using the email for the username" or something similar.

website/src/auth.tsx
```diff
@@ -64,6 +68,54 @@ export const UserProvider = (props: { children: any }) => {
...
+    user?.confirmRegistration(verificationCode, false, (err, result) => {
+      if (err) {
+        alert(err.message || JSON.stringify(err))
+        return
+      }
+      console.log(result)
+      setUser(user)
+      navigate("/auth/login") 
+    })
+  }
```
CharlieMcVicker, line 117: Something here feels weird to me but if this works I'm sure it's for a reason. There seem to be some weird things about verified users we may want to shelter the rest of the frontend from. Based on this code it seems I can be "logged in" and `useUser` will get me the user I'm logged in as **even if I still haven't verified my email** - that feels ungood.

Also this `setUser(user)` sort of scares me. Does this flow work without this? Why not?

website/src/auth.tsx
```diff
@@ -135,7 +188,14 @@ export const UserProvider = (props: { children: any }) => {
       value={{
         user,
         setUser,
```
CharlieMcVicker, line 190: Two things:
1. Should we do something here where we only expose the user to consumers if their email is verified?
2. Should we add a property to this hooks return like `unverifiedUser` that we set instead? That way pages that specifically need to handle users that have not verified their email yet still can do so.
3. Do we actually want to expose `setUser` to consumers? I would feel better if we knew that only this code here was allowed to `setUser` and only via the exposed operations.

website/src/pages/auth/confirm-signup.page.tsx
```diff
@@ -0,0 +1,58 @@
...
+} from "reakit"
+import { useUser } from "src/auth"
+import {
+  FormFields,
+  FormSubmitButton,
+  UserAuthPageTemplate,
+} from "./user-auth-layout"
+import { centeredForm } from "./user-auth.css"
+
+const VerificationPage = () => {
```
CharlieMcVicker, line 14: How do I ever get back to this page if I close the tab I started this workflow in? Is it included in the email?

CharlieMcVicker: Also your file and main component are fighting over what this page is called. Are we confirming the sign up or are we verifying the account?

website/src/auth.tsx
```diff
@@ -64,6 +68,54 @@ export const UserProvider = (props: { children: any }) => {
...
+
+  function verifyUser(verificationCode: string) {
+    user?.confirmRegistration(verificationCode, false, (err, result) => {
+      if (err) {
+        alert(err.message || JSON.stringify(err))
+        return
+      }
+      console.log(result)
+      setUser(user)
+      navigate("/auth/login")
```
CharlieMcVicker, line 115: Does the user have to sign in again after verifying? If so, why do we call `setUser` above here?

website/src/auth.tsx
```diff
@@ -82,6 +134,7 @@ export const UserProvider = (props: { children: any }) => {
 
         console.log("Login success. Result: ", data)
         alert("Login successful")
+        navigate("/")// TODO navigate to dashboard
```
CharlieMcVicker, line 137: Perhaps we need to fork here based on if the user is verified and possibly send them to the `auth/confirm-signup` page if they are not.

CharlieMcVicker (changes requested):
I think we have some gaps in the verification workflow.
1. If I create an account, close the tab, check my email, and get my token --> Can I still verify? How do I get to that verification page?
2. Are we letting unverified users masquerade as fully set up users on any page that is too lazy to check `emailVerified` (currently all pages)?  That doesn't feel good.

website/src/pages/auth/user-auth-layout.tsx
GracefulLemming: So, according to the actual HTML specs, this is technically an appropriate use of label. That being said, Web Content Accessibility Guidelines (WCAG) 2.0 would disallow using `label` here. I will certainly follow the more restrictive guidelines here, especially since accessibility is impacted. Change also made for `ResetLink` and `LoginLink`.

website/src/auth.tsx
GracefulLemming: AWS will give us a `UserNotConfirmedException` if a user tries to login but their account is not confirmed. Ive also added some nicer error handling for when this error does appear.

Relevant docs: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html#API_InitiateAuth_Errors

website/src/auth.tsx
GracefulLemming: For (1) and (2), the AWS Cognito API alerts us to unverified users where it matters (most notably, for `GetUser` and `InitiateAuth`. Since our group and auth guards both require access to the JWT, they'll fail if we get a `UserNotConfirmedException`. Ive also added general handling for when we get back a `UserNotConfirmedException` that will alert the user and redirect to confirmation.

Implemented (3).

CharlieMcVicker (changes requested):
A couple of things I found while (admittedly) trying to be a bit difficult.

1. I got a label but no button for `Resend confirmation code`
2. If I close the confirmation window and reopen, I am not prompted to sign back in (but I think I should be, right? We need my email or something?)
  a. Maybe we just need big text that says `Do not close this page`
  b. Maybe we should redirect from this page if we don't have a user email address available in state
4. It looks like maybe this is an `a` with `onClick` (ick), but I see no network activity when I click it.
5. The confirmation email doesn't say that it is for DAILP which might make it harder to find in my inbox.

Also, I will say, I think a discrete union for the auth state might be good. This would be more work, but would make us (me) very happy with our level of auth rigor going forward.

```ts
type UserContextType = NoAuth | NeedsVerification | SignedIn;

type NoAuth = {
  type: "NO_AUTH"
  user: null
  operations: {
    createUser: (username: string, password: string) => void
    loginUser: (username: string, password: string) => void
    resetPassword: (username: string) => void
  }
}

type NeedsVerification = {
  type: "NEEDS_VERIFCATION"
  user: null
  userProvidedEmail: string // not nullable
  operations: {
    resetConfirmationCode: () => void
    confirmUser: (confirmationCode: string) => void
  }
}


type SignedIn =  {
  type: "SIGNED_IN"
  user: CognitoUser // no longer nullable
  operations: {
    signOut: () => void
    changePassword: (verificationCode: string, newPassword: string) => void
  }
```

As an alternative to the below, you could split `operations` into its own chunk, where you define things like the below. Then you have a `useState<AuthState>` hook which will enforce that `setState` always keeps state valid.

```ts
type AuthState = SignedInState | ...;
type AuthContext = SignedInContext | ...;

type SignedInState =  {
  type: "SIGNED_IN"
  user: CognitoUser // no longer nullable
}

type SignedInContext = SignedInState & {
  operations: {
    signOut: () => void
    changePassword: (verificationCode: string, newPassword: string) => void
  }
}
```

GracefulLemming:
@CharlieMcVicker I wanted to give you an outline of the fixes I will be making since youre mostly async this week:
1. Implement fixes for your (1-3) in this branch
2. Make a ticket for (4) which requires changes in Terraform that would needlessly clutter this branch
3. Make a ticket "Use discrete union for auth state" so we can spend some time concretely planning out this work

GracefulLemming:
Jira issues:
https://dailp.atlassian.net/browse/DP-287
https://dailp.atlassian.net/browse/DP-286

website/src/auth.tsx
```diff
@@ -59,22 +59,26 @@ export const UserProvider = (props: { children: any }) => {
...
-          confirm(`This confirmation code has expired. Request a new code?`)
-        ) {
-          resetConfirmationCode()
+        let userResponse = confirm(
+          `This confirmation code has expired. Request a new code?`
+        )
+        if (userResponse && userProvidedEmail) {
+          resetConfirmationCode(userProvidedEmail)
+        } else if (userResponse) {
+          alert("We are unable to send a new code at this time.")
```
CharlieMcVicker, line 68: Perhaps we should not alert the user that they can request a new code if they haven't given us their email.

CharlieMcVicker (approved):
Provided reset condition is cleaned up, this looks all good!

PR #360, opened 2023-11-14
Enforce-code-styling

GracefulLemming:
Added formatting checks for rust and typescript code

In the future we may want to enforce:
- Formatting/linting of sql code using `sqlfluff format` or `sqlfluff lint`. Linting would require ignoring certain parsing errors, format would involve upgrading sqlfluff. Both will take a bit of research.
- Formatting of nix code using `nix fmt`
- Adequate documentation of rust (and maybe TS) code. This would resolve _many_ rust build warnings.

### DIFF
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -34,7 +34,6 @@ jobs:
       - name: Check code quality
         run: |
           echo "::group::Checking Rust format"
-          cargo install rustfmt
           cargofmt="$(cargo fmt --check)"
           if [[ -n $cargofmt ]]; then
             echo $cargofmt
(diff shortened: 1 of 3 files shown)


.github/workflows/terraform-plan.yml
```diff
@@ -31,6 +31,19 @@ jobs:
...
+          rustfmt = "$(cargo fmt --check)"
+          echo $rustfmt
+          echo "::endgroup"
+          if [[ -n $rustfmt ]]; then
+            echo "::error::Rust code does not meet code standards. Please run `cargo fmt` or `dev-check`."
+          fi
+          echo "::group::Checking format for frontend code"
+          cd website
+          yarn prettier -c --config package.json src
+          echo "::endgroup"
```
CharlieMcVicker, line 46: How does this lead to CI failing? Do `cargo fmt --check` and `yarn prettier -c` have non-zero exits that bubble up or something?

CharlieMcVicker (commented):
Looks like this is still WIP since it fails on this branch but overall looks like what I'd expect based on what we talked about!

hazelyn11 (approved):

PR #361, opened 2023-11-14
[DP-241] Comments Front End

hazelyn11:
This is ready for review! [Link to Jira ticket](https://dailp.atlassian.net/browse/DP-241?atlOrigin=eyJpIjoiNjQzYzEwMTFiMGE1NGE3OGJjOGY4MDc4NjhlMzljNjYiLCJwIjoiaiJ9)

There are a few code things I did that are probably janky. I have left comments pointing out these spots-- advice on a better strategy in those areas especially would be more than welcome!

### DIFF
--- a/website/src/comment-panel.tsx
+++ b/website/src/comment-panel.tsx
@@ -1,0 +1,78 @@
+import React, { ReactNode } from "react"
+import * as Dailp from "src/graphql/dailp"
+import * as css from "./panel-layout.css"
+import { TranslatedParagraph } from "./segment"
+import { useState } from "react"
+
+export const CommentPanel = (p: {
+    word: Dailp.FormFieldsFragment | null
+    segment: Dailp.DocumentParagraph | null
+}) => {
+    const [newCommentText, setNewCommentText] = useState<string>('')
+    const [newCommentType, setNewCommentType] = useState('')
+
+    // Prob messing up smth here
+    const [postCommentResult, postComment] = Dailp.usePostCommentMutation()
+
+    const options = [
+        // There is def a better way to do this..
+        'Story',
+        'Suggestion',
+        'Question',
+      ];
+    
+    /** Call the backend GraphQL mutation. */
+    const runUpdate = async (variables: {
+        input: Dailp.PostCommentInput
+    }) => {
+        return await postComment(variables)
+    }
+
+    const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setNewCommentText(event.target.value);
+    };
+
+    const handleSelectChange = (event: React.ChangeEvent<HTMLOptionElement>) => {
+        setNewCommentType(e.target.value);
+    };
+
+    const handleSubmit = (e: React.MouseEvent<HTMLButtonElement>) => {
+        e.preventDefault()
+    
+        runUpdate({
+            input: {
+              parentId: p.word? p.word.id : (p.segment? p.segment.id : null),
+              parentType: p.word ? Dailp.CommentParentType.Word : Dailp.CommentParentType.Paragraph,
+              textContent: newCommentText,
+              commentType: newCommentType == 'Story' ? Dailp.CommentType.Story : 
+              (newCommentType == 'Suggestion' ? Dailp.CommentType.Suggestion 
+              : Dailp.CommentType.Question),
+            },
+          })
+    
+        console.log('Submitted!');
+    };
+    
+    return(<div>
+        <input
+          type="text"
+          placeholder="Add a comment"
+          value={newCommentText}
+          onChange={handleInputChange}
+        />
+        <div>
+      <label htmlFor="dropdown">Select an option:</label>
+      <select id="dropdown" value={newCommentType} onChange={handleSelectChange}>
+        <option value="">Select...</option>
+        {options.map((option, index) => (
+          <option key={index} value={option}>
+            {option}
+          </option>
+        ))}
+      </select>
+    </div>
+        <button type="submit" onClick={handleSubmit}>Post</button>
+      </div>)
+}
+
+export default CommentPanel
--- a/website/src/comment-panel.tsx
+++ b/website/src/comment-panel.tsx
@@ -1,8 +1,9 @@
 import React, { ReactNode } from "react"
 import * as Dailp from "src/graphql/dailp"
-import * as css from "./panel-layout.css"
 import { TranslatedParagraph } from "./segment"
 import { useState } from "react"
+import { Button } from "./components"
+import * as css from "./comment-panel.css"
 
 export const CommentPanel = (p: {
     word: Dailp.FormFieldsFragment | null
@@ -41,7 +42,7 @@ export const CommentPanel = (p: {
     
         runUpdate({
             input: {
-              parentId: p.word? p.word.id : (p.segment? p.segment.index : null),
+              parentId: p.word? p.word.id : (p.segment? p.segment.id : null),
               parentType: p.word ? Dailp.CommentParentType.Word : Dailp.CommentParentType.Paragraph,
               textContent: newCommentText,
               commentType: newCommentType == 'Story' ? Dailp.CommentType.Story : 
@@ -71,7 +72,7 @@ export const CommentPanel = (p: {
         ))}
       </select>
     </div>
-        <button type="submit" onClick={handleSubmit}>Post</button>
+        <Button type="button" className={css.commentButton} onClick={handleSubmit}>Save</Button>
       </div>)
 }
 
--- a/website/src/comment-panel.tsx
+++ b/website/src/comment-panel.tsx
@@ -11,17 +11,17 @@ export const CommentPanel = (p: {
   setCommentsPanel: React.Dispatch<SetStateAction<boolean>>
 }) => {
   const [newCommentText, setNewCommentText] = useState<string>("")
-  const [newCommentType, setNewCommentType] = useState("")
+  const [newCommentType, setNewCommentType] = useState<string>("STORY")
 
   // Prob messing up smth here
   const [postCommentResult, postComment] = Dailp.usePostCommentMutation()
 
-  const options = [
-    // There is def a better way to do this..
-    "Story",
-    "Suggestion",
-    "Question",
-  ]
+  const commentTypeNames: Record<string, Dailp.CommentType> = {
+    // ... TS will then make sure you have an entry for everything on the "CommentTag" type that you import from the codegen
+    Story: Dailp.CommentType.Story,
+    Suggestion: Dailp.CommentType.Suggestion,
+    Question: Dailp.CommentType.Question,
+  }
 
   /** Call the backend GraphQL mutation. */
   const runUpdate = async (variables: { input: Dailp.PostCommentInput }) => {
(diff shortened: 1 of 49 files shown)


website/src/comment-panel.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setCommentsPanel: React.Dispatch<SetStateAction<boolean>>
+}) => {
+    const [newCommentText, setNewCommentText] = useState<string>('')
+    const [newCommentType, setNewCommentType] = useState('')
+
+    // Prob messing up smth here
+    const [postCommentResult, postComment] = Dailp.usePostCommentMutation()
+
+    const options = [
+        // There is def a better way to do this..
```
hazelyn11, line 20: Is there a better way to do this than hard-coding it? Probably, right?

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as Dailp from "src/graphql/dailp"
+import { TranslatedParagraph } from "./segment"
+import { useState } from "react"
+import { Button } from "./components"
+import * as css from "./comment-panel.css"
+
+export const CommentPanel = (p: {
+    word: Dailp.FormFieldsFragment | null
+    segment: TranslatedParagraph | null
+    setCommentsPanel: React.Dispatch<SetStateAction<boolean>>
```
hazelyn11, line 11: The way I am doing this works, but also feels very iffy.

website/src/panel-layout.tsx
```diff
@@ -78,8 +87,39 @@ export const PanelLayout = (p: {
...
+  const handleComment = (e: React.MouseEvent<HTMLButtonElement>) => {
+    e.preventDefault()
+
+    if (commentsPanel === true) {
+      setCommentsPanel(false)
+    } else {
+      setCommentsPanel(true)
+    }
+  }
+
```
GracefulLemming, line 100: This if block can be cleaned up using the power of Boolean logic.

Heres what I think youre trying to do:
When commentsPanel is true, do setCommentsPanel(false).
When commentsPanel is false, do setCommentsPanel(true).

Lets see what would happen if we tried using negation here:
```TypeScript
setCommentsPanel(!commentsPanel)
```

What does this do? Lets find out:
```PlainText
1. assume commentsPanel is true
setCommentsPanel(!commentsPanel) (Given)
setCommentsPanel(!true) (substitute)
setCommentsPanel(false) (negation)

2. assume commentsPanel is false
setCommentsPanel(!commentsPanel) (Given)
setCommentsPanel(!false) (substitute)
setCommentsPanel(true) (negation)
```
Its a match! Please replace this whole if block with `setCommentsPanel(!commentsPanel)`.

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,103 @@
...
+            : Dailp.CommentType.Question,
+      },
+    })
+
+    alert("Your comment has been posted!")
+    console.log("Submitted!")
+    p.setCommentsPanel(false)
+  }
+
+  return (
```
GracefulLemming, line 63: Does our dependency `reakit` have tools to help you make this form? Reakit handles some accessibility stuff so we try to use it when we can. That being said, it can be a bit of a pain to work with, so I would love a second opinion from @CharlieMcVicker on this

GracefulLemming (changes requested):

website/src/comment-panel.tsx
GracefulLemming: left a comment about this elsewhere. To start, try
```
export const CommentPanel = (p: {
    parentText: Dailp.FormFieldsFragment | TranslatedParagraph  | null,
    setCommentsPanel: React.Dispatch<SetStateAction<boolean>>
```

website/src/comment-panel.tsx
GracefulLemming: see https://github.com/NEU-DSG/dailp-encoding/pull/361#discussion_r1414326461

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,103 @@
...
+  setCommentsPanel: React.Dispatch<SetStateAction<boolean>>
+}) => {
+  const [newCommentText, setNewCommentText] = useState<string>("")
+  const [newCommentType, setNewCommentType] = useState("")
+
+  // Prob messing up smth here
+  const [postCommentResult, postComment] = Dailp.usePostCommentMutation()
+
+  const options = [
+    // There is def a better way to do this..
```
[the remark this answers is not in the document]
CharlieMcVicker: Depending on what GQL codegen is passing off another option is to make a
```ts
const commentTagNames: Record<CommentTag, string> = {
  // ... TS will then make sure you have an entry for everything on the "CommentTag" type that you import from the codegen
  STORY: "Story",
}
```

Then you can map over that with `Object.entries(commentTagNames.map)`

website/src/comment-panel.tsx
CharlieMcVicker: I think we could use `reakit` but in my mind, there will already be good accessibility coverage if we just put everything in a plain `form` and change the button click to an `onSubmit` on the form with `preventDefault.` To be clear, I think that the `textarea`, `label`s, `select`s and submit `button` should go in that form, but probably not your header.

website/src/graphql/dailp/queries.graphql
```diff
@@ -497,3 +498,16 @@ mutation DeleteContributorAttribution(
...
+    ... on AnnotatedForm {
+      __typename
+      id
+    }
+    ... on DocumentParagraph {
+      __typename
+      id
+    }
+  }
+}
```
CharlieMcVicker, line 513: _When you do the comments read/delete stuff_ you will want to make sure this query resolves the `comments` field with a sensible fragment so that our frontend will have up-to-date data when we get the mutation response, and we don't have to refetch a comment query or something.

website/src/panel-layout.tsx
```diff
@@ -52,6 +59,8 @@ export const PanelLayout = (p: {
     variables: { system: cherokeeRepresentation },
   })
 
+  const [commentsPanel, setCommentsPanel] = useState(false)
```
CharlieMcVicker, line 62: I think for consistency with editing you might call this `isCommenting, setIsCommenting`

website/src/panel-layout.tsx
```diff
@@ -143,6 +183,20 @@ export const PanelLayout = (p: {
   return (
     <div className={css.wordPanelContent}>
       <>{panel}</>
+      {commentsPanel === false && (
+        <Button type="button" onClick={handleComment}>
```
CharlieMcVicker, line 187:
```suggestion
        <Button type="button" onClick={() => setCommentsPanel(!commentsPanel}>
```

CharlieMcVicker: This looks like it should only ever open the comments panel, right? Or this button shared between the open/closed states somehow?

website/src/panel-layout.tsx
```diff
@@ -143,6 +183,20 @@ export const PanelLayout = (p: {
   return (
     <div className={css.wordPanelContent}>
       <>{panel}</>
+      {commentsPanel === false && (
```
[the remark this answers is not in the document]
CharlieMcVicker: I also like a "ternary" (three term expression) for this.

```
isCommenting ? (<ComponentForWhenTrue/>) : (<ComponentForWhenFalse/>)
```

Tbh I prefer this to using a mix of `&&` and `||`

CharlieMcVicker (changes requested):
Looks pretty good!

The must-haves for me are:
1. Wrap your form components in an HTML `form` element
2. Try to clean up the `commentTag` options stuff, somehow
3. Think about naming consistency with the `isEditing` flag
4. Try to clean up some of the props / onClick hooks and stuff (@GracefulLemming's comments)

Great job! These are sturdy bones to improve upon.

website/src/comment-panel.tsx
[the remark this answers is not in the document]
GracefulLemming: it seems `map()` does not exist on `Record`

hazelyn11:
@CharlieMcVicker @GracefulLemming  Thank you for all of your helpful comments! I just pushed changes which should completely address all of your comments-- I have a few questions about ways I fixed certain things and will leave those as comments in my code

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,108 @@
...
+  segment: Dailp.FormFieldsFragment | TranslatedParagraph
+  setIsCommenting: React.Dispatch<SetStateAction<boolean>>
+}) => {
+  const [newCommentText, setNewCommentText] = useState<string>("")
+  const [newCommentType, setNewCommentType] = useState<string>("STORY")
+
+  // Prob messing up smth here
+  const [postCommentResult, postComment] = Dailp.usePostCommentMutation()
+
+  const commentTypeNames: Record<string, Dailp.CommentType> = {
```
hazelyn11, line 18: @CharlieMcVicker Is this what you were envisioning?

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <label htmlFor="dropdown" className={css.spacing}>
+            Tag:
+          </label>
+          <select
+            id="dropdown"
+            value={newCommentType}
+            onChange={handleSelectChange}
+            className={css.spacing}
+          >
+            {Object.entries<Dailp.CommentType>(commentTypeNames).map(
```
hazelyn11, line 87: @CharlieMcVicker Re: is this what you were envisioning for storing and displaying the commentType names? I will say while implementing this, it felt substantially more complicated and less intuitive for any improvement in functionality that I could understand-- since I think we would still have to go manually update the Record if we wanted to at some point add more commentTypes. I also could have just misunderstood what you were suggesting, though

website/src/comment-panel.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <div>
+      <header className={css.wordPanelHeader}>
+        <h2 className={css.editCherHeader}>
+          {p.segment.__typename === "DocumentParagraph"
+            ? "Paragraph " + p.segment?.index
+            : p.segment.source}
+        </h2>
+      </header>
+      <form>
```
hazelyn11, line 70: Is adding this form tag here sufficient?

website/src/comment-panel.tsx
CharlieMcVicker: This is what I was thinking (expect that `[label, option]` will flip when you flip the record). I think that even though we **still have to go manually update the Record**, that's actually necessary. If somebody adds a new tag, they shouldn't have to assume that what they put in that enum is going to be directly displayed to a user. That's not what it means to be in an enum, and so we can't assume that someone editing the enum knows anything about how stuff gets displayed. And we actually probably don't want them to have to think about that.

Now, if someone makes a change to the enum (ie. they add a tag), they will get an error that says "This object called `commentTypeNames` promised to have a user-facing name for every comment type, but it's missing a field for _your new type_." I think that should pretty quickly point them in the direction of "Oh I should add a user facing name." IMO that's sort of the best we can hope for in this case.

If we did something clever like map the user-facing names from the enum values we could have something happen like:
1. A developer adds a new tag with some weird developer facing name `COMMENT_SUGGESTION_REPLY_V2` or something. Idk people name stuff weird
2. The code compiles fine and so it gets merged
3. Somebody encounters this weird developer speak and issues a bug
4. We notice the names are mapped from the enum and so we'd have to change the enum
5. Values of `COMMENT_SUGGESTION_REPLY_V2` are already in the database so we have also have to run a database migration to change `COMMENT_SUGGESTION_REPLY_V2` to something more human-readable
6. Three weeks later, we decide we don't like the name we picked and we want to change it again, and we have to update the schema and run a migration, again.

website/src/panel-layout.tsx
```diff
@@ -183,19 +158,18 @@ export const PanelLayout = (p: {
...
-          onClick={handleComment}
+          onClick={() => setIsCommenting(!isCommenting)}
           className={css.buttonSpacing}
         >
           Discard
         </SubtleButton>
+      ) : (
+        <Button type="button" onClick={() => setIsCommenting(!isCommenting)}>
+          Comment
+        </Button>
```
CharlieMcVicker, line 172: Does this need to also post the comment?

CharlieMcVicker (changes requested):
I think that record needs to be flipped but this is super close. I left some other small things too, and responded to your comment on the `Object.entires` and `map` stuff.

website/src/comment-panel.tsx
hazelyn11: Makes sense to me! I flipped it

website/src/comment-panel.tsx
hazelyn11: Thanks so much for this explanation! I understand now and this makes sense to me

website/src/panel-layout.tsx
hazelyn11: I don't believe so, the comment is posted within the CommentPanel component!

CharlieMcVicker: OH this opens the panel I get it now

CharlieMcVicker (approved):
Woohoo! Looks good, and works on my local.

GracefulLemming (approved):

PR #370, opened 2023-12-14
[DP-276] Comments Viewing FrontEnd

hazelyn11:
[Jira ticket
](https://dailp.atlassian.net/browse/DP-276)

Now we can view comments both in the word pane and the paragraph pane!

One note-- in order to render comments immediately after they are submitted, I am currently also showing the collapsible Discussion panel even when it is empty. Otherwise, the pane has to be closed and re-opened in order for comments to show. Maybe there's a better way around this?

image

### DIFF
--- a/website/src/components/comment-section.css.ts
+++ b/website/src/components/comment-section.css.ts
@@ -1,9 +1,7 @@
-import { style } from "@vanilla-extract/css"
-import { withSSRContext } from "aws-amplify"
-import { borderWidth, margin } from "polished"
+import { style, styleVariants } from "@vanilla-extract/css"
+import { margin } from "polished"
 import {
   colors,
-  fonts,
   hspace,
   mediaQueries,
   radii,
--- a/website/src/components/comment-section.css.ts
+++ b/website/src/components/comment-section.css.ts
@@ -1,0 +1,83 @@
+import { style } from "@vanilla-extract/css"
+import { withSSRContext } from "aws-amplify"
+import { borderWidth, margin } from "polished"
+import {
+  colors,
+  fonts,
+  hspace,
+  mediaQueries,
+  radii,
+  tagColors,
+  thickness,
+  vspace,
+} from "src/style/constants"
+import { marginX, marginY, paddingX, paddingY } from "src/style/utils"
+
+const wordShared = style([
+  paddingY(vspace.quarter),
+  paddingX(hspace.halfEdge),
+  {
+    borderWidth: thickness.thick,
+    borderStyle: "solid",
+    "@media": {
+      [mediaQueries.print]: {
+        ...margin(0, "3.5rem", vspace[1.5], hspace.small),
+      },
+      [mediaQueries.medium]: margin(0, "2.5rem", vspace.one, hspace.small),
+    },
+  },
+])
+
+export const commentWrapper = style([
+  wordShared,
+  marginY(vspace.half),
+  {
+    borderColor: colors.borders,
+    borderRadius: radii.large,
+    lineHeight: vspace.one,
+    pageBreakInside: "avoid",
+    breakInside: "avoid",
+    "@media": {
+      [mediaQueries.medium]: {},
+    },
+  },
+])
+
+export const headerStyle = style([
+  marginX(hspace.medium),
+  {
+    fontSize: "0.7rem",
+  },
+])
+
+export const tagPadding = style([paddingY(vspace.medium)])
+
+export const tagColorStory = style([
+  paddingX(hspace.medium),
+  {
+    display: "inline-block",
+    borderRadius: radii.round,
+    backgroundColor: tagColors.story,
+    fontSize: "0.7rem",
+  },
+])
+
+export const tagColorSuggestion = style([
+  paddingX(hspace.medium),
+  {
+    display: "inline-block",
+    borderRadius: radii.round,
+    backgroundColor: tagColors.suggestion,
+    fontSize: "0.7rem",
+  },
+])
+
+export const tagColorQuestion = style([
+  paddingX(hspace.medium),
+  {
+    display: "inline-block",
+    borderRadius: radii.round,
+    backgroundColor: tagColors.question,
+    fontSize: "0.7rem",
+  },
+])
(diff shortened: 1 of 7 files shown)


website/src/components/comment-section.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <div className={css.commentWrapper}>
+      {p.comment.textContent}
+      <div className={css.tagPadding}>
+        <div
+          className={
+            p.comment.commentType === Dailp.CommentType.Story
+              ? css.tagColorStory
+              : p.comment.commentType === Dailp.CommentType.Suggestion
+              ? css.tagColorSuggestion
+              : css.tagColorQuestion
```
GracefulLemming, line 83: One thing to consider: You can have named style variants that share some common styling but diverge based on a parameter. Using a variant might clean up things here and lead to less redundant css.

read more: https://vanilla-extract.style/documentation/api/style-variants/

website/src/components/comment-section.css.ts
```diff
@@ -0,0 +1,83 @@
+import { style } from "@vanilla-extract/css"
+import { withSSRContext } from "aws-amplify"
```
GracefulLemming, line 2: be sure to remove unused imports

website/src/components/comment-section.tsx
```diff
@@ -0,0 +1,102 @@
...
+  parent: Dailp.FormFieldsFragment | TranslatedParagraph
+}) => {
+  return (
+    <div>
+      {p.parent.__typename === "DocumentParagraph" ? (
+        <ParagraphCommentSection paragraph={p.parent} />
+      ) : p.parent.__typename === "AnnotatedForm" ? (
+        <WordCommentSection word={p.parent} />
+      ) : (
+        ""
```
GracefulLemming, line 17: `parent` cannot be `undefined`, `null`, or some other type. You can delete this embedded ternary to leave you with the following:

```
{p.parent.__typename === "DocumentParagraph" ? (
         <ParagraphCommentSection paragraph={p.parent} />
       ) : <WordCommentSection word={p.parent} />)}

File: website/src/components/comment-section.tsx [hunk 4]
```diff
@@ -0,0 +1,102 @@
...
+    </div>
+  )
+}
+
+export const WordCommentSection = (p: { word: Dailp.FormFieldsFragment }) => {
+  const [{ data }] = Dailp.useWordCommentsQuery({
+    variables: { wordId: p.word.id },
+  })
+
+  const wordComments = data?.wordById
```
GracefulLemming, line 28: if youre naming this `wordComments`, then maybe it should hold the actual list of comments and not the word info.

```
let wordComments = data?.wordById?.comments
```

Then line 32 would read

```
wordComments?.map(...)
```

website/src/components/comment-section.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export const ParagraphCommentSection = (p: {
+  paragraph: TranslatedParagraph
+}) => {
+  const [{ data }] = Dailp.useParagraphCommentsQuery({
+    variables: { paragraphId: p.paragraph.id },
+  })
+
+  const paragraphComments = data?.paragraphById
```
GracefulLemming, line 49: same notes as the word version

GracefulLemming (changes requested):

website/src/components/comment-section.tsx
hazelyn11: when I try to delete the embedded ternary I get this error: ```
Type 'FormFieldsFragment | ({ readonly __typename?: "DocumentParagraph" | undefined; } & Pick<DocumentParagraph, "id" | "index" | "translation"> & { ...; })' is not assignable to type 'FormFieldsFragment'.
  Type '{ readonly __typename?: "DocumentParagraph" | undefined; } & Pick<DocumentParagraph, "id" | "index" | "translation"> & { readonly source: readonly (({ readonly __typename: "AnnotatedForm"; } & Pick<...> & { ...; }) | { ...; })[]; readonly comments: readonly ({ ...; } & Pick<...>)[]; }' is not assignable to type 'FormFieldsFragment'.
    Type '{ readonly __typename?: "DocumentParagraph" | undefined; } & Pick<DocumentParagraph, "id" | "index" | "translation"> & { readonly source: readonly (({ readonly __typename: "AnnotatedForm"; } & Pick<...> & { ...; }) | { ...; })[]; readonly comments: readonly ({ ...; } & Pick<...>)[]; }' is not assignable to type '{ readonly __typename?: "AnnotatedForm" | undefined; }'.
      Types of property '__typename' are incompatible.
        Type '"DocumentParagraph" | undefined' is not assignable to type '"AnnotatedForm" | undefined'.
```

File: website/src/components/comment-section.tsx (hunk 1 above)
Comment (hazelyn11) [reply website/src/components/comment-section.tsx:83]: Hmm, I'm trying this and it's fine to define a variant but I'm having trouble actually passing in the values I need to use it from the component without getting errors. If this is essential I can keep trying to figure out how to do it but I've spent a while on it and am struggling

**Comment by hazelyn11** [comment]
@GracefulLemming I have addressed most of your comments and left follow-ups on two of them, lmk your thoughts!

File: website/src/components/comment-section.tsx (hunk 3 above)
Comment (CharlieMcVicker) [reply website/src/components/comment-section.tsx:16]: It looks like typescript is not being nice to you. Sometimes nested fields will be like this when you try to do a type guard (checking if a value is of a certain type, like you do when you check the value of `__typename`) on a literal. You might see if this goes away if instead of taking in an object of params like this:
```
(p: {...prop type}): ReactElement
```
You instead "destructure" your props like this:
```
({parent}: {... prop type}): ReactElement
```

Here are some docs on "destructuring" in JS if that is new for you: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

CharlieMcVicker: We may also fix this if we upgrade typescript somehow. This should work.

CharlieMcVicker (approved):
Up to @GracefulLemming for final approval but this looks good to me.

hazelyn11:
> Up to @GracefulLemming for final approval but this looks good to me.

thanks for your comments Charlie!

GracefulLemming (approved):

GracefulLemming:
@hazelyn11 if you want to finish the suggestions Charlie made, feel free to. Otherwise,  you can merge!

PR #374, opened 2023-12-19
AWS ApiGateway will encode user groups as a comma-separated string

CharlieMcVicker:
Jira: https://dailp.atlassian.net/browse/DP-290?atlOrigin=eyJpIjoiYmZkN2EyNGI4YmRkNDgwMWI1YTEyYzVhNmI3YTY1NTYiLCJwIjoiaiJ9

# What this does and why
We are having odd behavior where users who are signed in are getting "Forbidden, user not authenticated" errors. This is because we were failing to decode the user's identity token. While digging into this, I uncovered some other inconsistencies we have to plan around. Namely, our JWTs have user groups as an Array of strings, while API gateway will always give us a comma-separated string. To make this work, I have decoupled the deserialization models from the `UserInfo` type.

## Changes
- `UserInfo` can no longer be deserialized directly, and callers have to use either `JWTUserInfo` or `ApiGatewayUserInfo`, which have slightly different ways of encoding user groups
- `UserInfo`, `UserGroup`, `AuthGuard` and `GroupGuard` now live in a new `auth` module in `types/`

## Future-proofing goodness
We are now able to add Guards to fields in `types/` if we ever need to. Ie. if only an editor should be able to retrieve a field on some type.


types/src/user.rs
```diff
@@ -48,5 +54,78 @@ pub enum UserGroup {
 // // Impl FromStr and Display automatically for UserGroup, using serde.
 // // This allows us to (de)serialize lists of groups via a comma-separated string
 // // like this: "Editor,Contributor,Translator"
-// serde_plain::forward_from_str_to_serde!(UserGroup);
-// serde_plain::forward_display_to_serde!(UserGroup);
+serde_plain::forward_from_str_to_serde!(UserGroup);
+serde_plain::forward_display_to_serde!(UserGroup);
```
CharlieMcVicker, line 58: @nole2701 in the future please avoid commenting out pieces of code! It looks like this was removed when you moved these declarations into the `types` codebase.

PR #381, opened 2023-12-29
Add-auth-buttons-to-collections

GracefulLemming:
There is no way to add auth navigation to a generic edited collection template (yet), so instead I just added it to the CWKW header.

Colors changed to better match CWKW style, but are open to change (as long as we are using CWKW colors)


CharlieMcVicker (approved):
LGTM

PR #382, opened 2023-12-29
Format queries/mutations for comments to work with GraphCache

CharlieMcVicker:
Jira: [Comments only sometimes work](https://dailp.atlassian.net/browse/DP-296?atlOrigin=eyJpIjoiYTllYTJmNGMzYTEzNDM4ZTkwNWJlYWNiN2M5ODNkNDgiLCJwIjoiaiJ9)

Currently, comments are posted successfully, but the UX doesn't update automatically. By changing what we fetch in the queries (there is no change to the typescript here) we will get the caching behavior we want. That is, when you post a comment, the word/paragraph comments query will be updated.

This PR should be a great example of making something work with GraphCache!

### DIFF
--- a/website/src/graphql/dailp/queries.graphql
+++ b/website/src/graphql/dailp/queries.graphql
@@ -427,48 +427,39 @@ query BookmarkedDocuments {
   }
 }
 
+fragment CommentFields on Comment {
+  id
+  postedAt {
+    timestamp
+    date {
+      year
+      month
+      day
+      formattedDate
+    }
+  }
+  postedBy {
+    id
+    displayName
+  }
+  textContent
+  commentType
+}
+
 query WordComments($wordId: UUID!) {
   wordById(id: $wordId) {
+    id
     comments {
-      id
-      postedAt {
-        timestamp
-        date {
-          year
-          month
-          day
-          formattedDate
-        }
-      }
-      postedBy {
-        id
-        displayName
-      }
-      textContent
-      commentType
+      ...CommentFields
     }
   }
 }
 
 query ParagraphComments($paragraphId: UUID!) {
   paragraphById(id: $paragraphId) {
+    id
     comments {
-      id
-      postedAt {
-        timestamp
-        date {
-          year
-          month
-          day
-          formattedDate
-        }
-      }
-      postedBy {
-        id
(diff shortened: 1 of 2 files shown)


website/src/graphql/dailp/queries.graphql
```diff
@@ -427,48 +427,39 @@ query BookmarkedDocuments {
...
+    id
+    displayName
+  }
+  textContent
+  commentType
+}
+
 query WordComments($wordId: UUID!) {
   wordById(id: $wordId) {
+    id
```
CharlieMcVicker, line 451: We always need to fetch IDs if we want this to work

website/src/graphql/dailp/queries.graphql
```diff
@@ -427,48 +427,39 @@ query BookmarkedDocuments {
...
+      formattedDate
+    }
+  }
+  postedBy {
+    id
+    displayName
+  }
+  textContent
+  commentType
+}
```
CharlieMcVicker, line 447: These common fields used by two queries and the post mutation are now in a fragment.

website/src/graphql/dailp/queries.graphql
```diff
@@ -427,48 +427,39 @@ query BookmarkedDocuments {
...
-        }
-      }
-      postedBy {
-        id
-        displayName
-      }
-      textContent
-      commentType
+      ...CommentFields
     }
```
CharlieMcVicker, line 454: The fragment replaces these fields here

website/src/graphql/dailp/queries.graphql
```diff
@@ -603,10 +594,16 @@ mutation PostComment($input: PostCommentInput!) {
...
+      comments {
+        ...CommentFields
+      }
     }
     ... on DocumentParagraph {
       __typename
       id
+      comments {
+        ...CommentFields
+      }
```
CharlieMcVicker, line 606: We fetch the comments and the fields we need when we post, so we have an up-to-date version on the frontend as soon as the mutation finishes.

CharlieMcVicker (commented):
Tour of changes

GracefulLemming (approved):

PR #387, opened 2024-01-16
Clean up the creation of GraphQL clients in our frontend

CharlieMcVicker:
- minimize calls to createClient
- clean up names and move all createClient calls to one file
- refactor(gql): auth exchange decides which gql endpoint to use

In light of issues Ellen has had and Naomi has reproduced, it was time to simplify all the code creating GraphQL clients in our codebase. Instead of swapping out clients based on if the user has a token, we now have a smarter `AuthExchange` that ensures we always include a token for authenticated users (and hit `/graphql-edit`) and that we do not include tokens for un-authenticated users and WordPress GraphQL requests.

I think there is probably still some refining and renaming to do on this, specifically the `serverSideClients` could be better named.


website/src/graphql/authExchange.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (getCurrentUser()) {
+        // if we are signed in, we must have a valid token
+        return !authState || isExpired(authState.token)
+      } else if (authState) {
+        // we aren't signed in but we have a token, we need to get rid of it
+        return true
+      } else {
+        return false
+      }
+    },
```
CharlieMcVicker, line 37: Let's make sure this is really correct

website/src/graphql/authExchange.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  GRAPHQL_URL_READ,
+  GRAPHQL_URL_WRITE,
+  WP_GRAPHQL_URL,
+} from "src/graphql"
+
+const secondsSinceEpoch = () => Date.now() / 1000
+
+const isExpired = (token: CognitoIdToken) =>
+  token.getExpiration() <= secondsSinceEpoch()
```
CharlieMcVicker, line 14: I think this is right. The API has no docstring for `getExpiration`

website/src/graphql/index.ts
```diff
@@ -0,0 +1,122 @@
...
+export const serverSideClients = {
+  dailp: createClient({
+    url: GRAPHQL_URL_READ,
+    exchanges: [sharedCache, fetchExchange],
+  }),
+  wordpress: createClient({
+    url: WP_GRAPHQL_URL,
+    exchanges: [sharedCache, fetchExchange],
+  }),
+}
```
CharlieMcVicker, line 78: I think these should be renamed but I don't have any clue what they should be.

website/src/graphql/authExchange.ts
CharlieMcVicker: I'm following this which is based on the JWT spec https://stackoverflow.com/questions/39926104/what-format-is-the-exp-expiration-time-claim-in-a-jwt

website/src/graphql/index.ts
```diff
@@ -0,0 +1,122 @@
+import {
```
CharlieMcVicker, line 1: this was actually moved from `graphql.ts` -> `graphql/index.ts`

CharlieMcVicker (commented):
Some things to look at.

website/src/graphql/authExchange.ts
GracefulLemming: I think something here isnt working as intended- I kept a window with a user logged in all weekend and the token did not expire.

GracefulLemming (changes requested):

website/src/graphql/authExchange.ts
CharlieMcVicker: When you say that, do you mean the token was properly refreshed, or that we never detected the expiry and kept sending a bad token?

GracefulLemming (approved):

PR #389, opened 2024-01-18
return value of UserInfo is now optional

GracefulLemming:


CharlieMcVicker (approved):
oh dear lord yep let's not `_unchecked` anything if we can avoid it 😭

PR #390, opened 2024-01-19
Create-managed-signout-function

GracefulLemming:


website/src/auth.tsx
```diff
@@ -268,6 +269,11 @@ export const UserProvider = (props: { children: any }) => {
     })
   }
 
+  function signOutUser() {
+    setUser(null)
+    user?.signOut()
```
CharlieMcVicker, line 274: should we do the `setUser(null)` in a callback? In case signing out can error or something (ie. bad network)

CharlieMcVicker (changes requested):
I think we should be a little more careful here.

CharlieMcVicker (approved):

PR #391, opened 2024-01-19
Add-home-link-to-collections-header

GracefulLemming:


website/src/pages/cwkw/cwkw-layout.tsx
```diff
@@ -47,7 +47,7 @@ const CWKWLayout: React.FC = ({ children }) => {
           }
         >
           {isDesktop ? <Sidebar /> : <MobileSidebar />}
-          <div className={css.contentContainer}>
+          <span className={css.contentContainer}>
             <h1 className={css.siteTitle}>
```
CharlieMcVicker, line 51: This is unintuitive to me. A `span` with a `h1` and `p` in it is sort of odd, no? Also, we just make it `display: block` again anyways. If this is valid HTML and there is a reason behind it, I think it warrants a comment.

CharlieMcVicker (changes requested):

website/src/pages/cwkw/cwkw-layout.tsx
GracefulLemming: oops, that was just me getting my HTML semantics scrambled. Will fix!

website/src/pages/cwkw/cwkw-layout.css.ts
```diff
@@ -39,11 +39,31 @@ export const subHeader = style([baseLayout.subHeader, { fontSize: "larger" }])
...
+  },
+])
 
 export const contentContainer = style([
   baseLayout.contentContainer,
-  { flex: 1 },
+  {
+    display: "block",
+  },
 ])
```
CharlieMcVicker, line 67: This doesn't seem right to me. I think we wanted to keep the flex property and get rid of display block, since this is applied to a block level element?

CharlieMcVicker (approved):
approved pending css fixes

PR #393, opened 2024-01-22
Disable-readers-word-edit

GracefulLemming:


website/src/edit-doc-data-panel.tsx
```diff
@@ -106,6 +104,14 @@ export const EditDocPanel = (props: { document?: Dailp.AnnotatedDoc }) => {
     }
   }, [day, month, year])
 
+  /* TODO (NOTE): While we could disable the edit button for "Document Info" to non-editors,
+  in the near future we will want users with the Contributor and Editor roles
+  to be able to add people to the "Contributors" list in the document info.
+  We want the above functionality in order to support in-person work teams who prefer working
+  on paper or through discussion, then putting everything in the computer at once 
+  via one DAILP user account. */
+  let userRole = useUserRole()
```
CharlieMcVicker, line 113: (I think it's out of scope for this ticket but we should still think about it) Should this be done with groups and not with roles? Ie. the important thing is having a group (you may have others) not being a particular role (you only have one)? We haven't really specified how any of this should work, but I think it's worth raising for thought.

CharlieMcVicker (approved):
One point for discussion, but this looks good.

LN7-cyber:
Word panel tsx: Trim panel previews and make sure that word panel and paragraph panel options are in their appropriate places. Review stale code in the word panel and paragraph panel. Separate the common panel from the word panel file.

PR #397, opened 2024-01-26
Add instructions for bookmarking documents

GracefulLemming:
@LN7-cyber Could you and/or Milan review the new text on the bookmarks tab and edit as needed?


CharlieMcVicker (approved):
Looks good!

PR #399, opened 2024-01-30
add cloudfront url to amplify settings

GracefulLemming:
@LN7-cyber & @CharlieMcVicker:
This resolves the issue discussed in today's meeting.

@CharlieMcVicker: **Important** With the current code, amplify will have a url in the format `xxxxxxxxxxxxx.cloudflare.net/` and *not* `https://xxxxxxxxxxxxx.cloudflare.net/`. I can change this for simplicity though.


CharlieMcVicker (approved):
This format (without the protocol) is what the audio code expects to consume. Thanks so much!

PR #400, opened 2024-02-01
Create-post-confirmation-lambda-event-handler

GracefulLemming:
Creates a new rust member for housing lambda event handlers that are not a core part of the application.

Creates a new rust binary with logic for a lambda that adds users to a group if their email is on a predefined list when they confirm their account.


admin-event-handlers/src/auth/post-confirmation/cognito_idp_operations.rs
```diff
@@ -0,0 +1,36 @@
...
+    ) -> Result<(), anyhow::Error> {
+        self.client
+            .admin_add_user_to_group()
+            .user_pool_id(self.pool_id)
+            .username(email)
+            .group_name(group.to_string())
+            .send()
+            .await
+            .map_err(|e| anyhow::Error::new(e))
+            .map(|x| ())
```
CharlieMcVicker, line 34: Does this line map the result to the empty tuple? I think I see the usecase here and if this is the standard pattern I'm fine with it. It just throws me a bit.

admin-event-handlers/src/auth/post-confirmation/google_sheets_operations.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Result obtained directly from the raw Google sheet.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SheetResult {
+    /// Each element here represents one row.
+    pub values: Vec<Vec<String>>,
+}
+
+impl SheetResult {
+    /// Calls the Google Sheets API a number of times.
+    /// Deserializes the first successful result, then stops attempting.
```
CharlieMcVicker, line 25: Is this duplicate code with migrations stuff?

admin-event-handlers/src/auth/post-confirmation/google_sheets_operations.rs
```diff
@@ -0,0 +1,73 @@
...
+            sheet_id, api_key
+        ))
+        .await?;
+        Ok(response.json::<SheetResult>().await?)
+    }
+    /// Reads each line of the spreadsheet and encodes any roles defined in it.
+    pub fn into_permission_list(self) -> Result<Vec<UserPermission>> {
+        let mut sections: Vec<UserPermission> = Vec::new();
+        // First row is headers "Full Name" "Alt name" "DOB" "Role" "email"
+        for row in self.values.into_iter().skip(1) {
```
CharlieMcVicker, line 53: We could also assert this instead of skipping

admin-event-handlers/src/auth/post-confirmation/main.rs
```diff
@@ -0,0 +1,84 @@
...
+    let region_provider = RegionProviderChain::first_try(Region::new(region))
+        .or_default_provider()
+        .or_else(Region::new("us-east-1"));
+
+    let config = aws_config::defaults(BehaviorVersion::latest())
+        .region(region_provider)
+        .load()
+        .await;
+
+    let cognito_action = CognitoClient::new(&config)
```
CharlieMcVicker, line 66: I think it would be alright to inject the user pool id here and not hide that dependency on an environment variable in this `new` function

CharlieMcVicker (approved):
None of these are blocking, but here are some nice to haves/questions.

admin-event-handlers/src/auth/post-confirmation/google_sheets_operations.rs
GracefulLemming: To clarify, do you mean using `Macro std::assert` here?

admin-event-handlers/src/auth/post-confirmation/google_sheets_operations.rs
GracefulLemming: See #402

PR #419, opened 2024-07-09
Add password visibility toggle

GracefulLemming:


nole2701 (approved):
Looks good to me. Using state to keep track of visibility and input type with an onclick toggle.

PR #420, opened 2024-07-11
Add environment marker bar

GracefulLemming:
Adds a visual indicator for showing which site environment a user is accessing. Bar should not be present on production.


nole2701 (approved):
looks good to me

PR #422, opened 2024-07-11
Disable postlogin handler

GracefulLemming:
Fixes issue where meaningless error messages appear during account confirmation


nole2701 (approved):
looks good to me

PR #424, opened 2024-07-24
Limit-webcrawler-access

GracefulLemming:
Prevents search engines from finding dev and uat pages. Also blocks some known AI scrapers.


nole2701 (approved):
Looks good!

PR #433, opened 2024-11-13
migrate sql schema for contributed/edited audio on documents

CharlieMcVicker:
This PR contains the database migrations (and updates to documentation) to support user-contributed audio on documents.

On the whole, my approach was to replicate what we did with user-contributed audio on words, namely:
1. Add columns on the `document` table to handle "editing" (hiding/showing) audio ingested from sheets
2. Create a join table (`document_id`, `media_slice_id`) for user contributed audio on documents
    1. Like how we did for words, I left the index as a compound index on both keys, which doesn't make a ton of sense (why would the same slice ever be tied to two documents?). But I think having an index on `document_id` is good, and that is listed first in the compound index. Alternatively, we could make the `media_slice_id` the primary key on this table, enforcing a many (slices) to one (document) relationship.

These migrations assume we are going to be replicating the editor workflow for curating contributed audio at the document level. I didn't see that in the mocks though, so let's double check before we merge.

In graphql, we will cover up this implementation detail of which audio is ingested and which audio is contributed, like we did for words.

### DIFF
--- a/doc/database/documents.md
+++ b/doc/database/documents.md
@@ -12,16 +12,18 @@
 
 ## `document`
 
-| column           | type                     | description                                  |
-| ---------------- | ------------------------ | -------------------------------------------- |
-| `id`             | `uuid`                   | Primary key                                  |
-| `short_name`     | `text`                   | URL slug, natural key                        |
-| `title`          | `text`                   | Full title of the document                   |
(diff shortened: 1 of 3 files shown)


GracefulLemming (approved):

GracefulLemming (approved):
This looks great! I am approving but would like to know if you would prefer this merging after the SQLX issue is resolved.

PR #436, opened 2024-11-26
Dennis tickets

hellolol2016:
made changes to support translation in backend, based on ticket written by naomi. sorry cant be more descriptive about to go on plane.


types/queries/update_word.sql
```diff
@@ -8,6 +8,12 @@ update word set
         case
             when $3::text[] != '{}' then $3[1]
             else commentary
+        end,
+
+    english_translation =
```
GracefulLemming, line 13: Should be called `english_gloss` as defined in `types/migrations/20220205042039_init.sql`

GracefulLemming, line 13: Should be called english_gloss as defined in types/migrations/20220205042039_init.sql

types/queries/update_word.sql
```diff
@@ -8,6 +8,12 @@ update word set
...
+        case
+            when $2::text[] != '{}' and $2[1] is not null then $2[1]
```
GracefulLemming, line 15: Each number in this file corresponds to an input parameter. Notice that in `types/src/database_sql.rs:515:27` you call this query with 4 parameters, but only three are defined here:
```
let document_id = query_file!(
    "queries/update_word.sql",
    word.id,            /// $1
    &source as _,       /// $2
    &commentary as _,   /// $3
    &translation as _,  /// $4? Does not exist!
)
```
Try replacing the reference to `$2` with `$4` instead!

website/src/edit-word-form-context.tsx
```diff
@@ -63,6 +63,7 @@ export const FormProvider = (props: { children: ReactNode }) => {
             source: values.word["source"],
             commentary: values.word["commentary"],
             segments: updatedSegments,
+            translation: values.word["translation"],
           },
```
GracefulLemming, line 67: Another confusing naming on our part. Should be `values.word["englishGloss"]` instead of `values.word["translation"]`.

You may run into a type mismatch here. I am looking into it but am blocked by another issue were having.

GracefulLemming (changes requested):
Congrats on your first PR! I added some comments for you to address.

types/queries/update_word.sql
hellolol2016: i also renamed the instance of english_translation below like this 
```
    english_gloss=
        case
            when $4::text[] != '{}' and $4[1] is not null then $4[1]
            else english_gloss
```

website/src/edit-word-form-context.tsx
hellolol2016: I am getting a type error here.
```
Type 'readonly string[]' is not assignable to type 'InputMaybe<string>'.
```

types/src/database_sql.rs
```diff
@@ -237,7 +237,6 @@ impl Database {
         .await?;
         Ok(words
             .into_iter()
-            .group_by(|w| (w.document_id, w.is_reference))
```
GracefulLemming, line 240: Why are we removing this line?

GracefulLemming (changes requested):

website/src/edit-word-form-context.tsx
GracefulLemming: I can take a look at this since it seems like this will be annoying to debug without intellisense helping out.

PR #437, opened 2024-12-04
Fix-dev-check-always-fails

GracefulLemming:
# *Issue*: `dev-check` always fails during backend checks.

## Background
The primary issue is that`dev-check` fails in all conditions and suggests running `cargo check` during backend checks. Running `cargo-check` as suggested finishes successfully and yields no further information. A closer look at `check.sh` reveals that `cargo check` only fails once `$SQLX_OFFLINE` is true, suggesting that the cause of this issue is upstream in the check script. The only upstream invocation of sqlx is in the database check phase, as shown below: 
``` bash
echo "Checking if SQL types are up to date..."
if $(cd types && cargo sqlx prepare --check -- -p dailp &>/dev/null); then
    echo "Generating SQL types..."
    cargo sqlx prepare -- -p dailp || exit 1
fi
```
Running `cargo sqlx prepare -- -p dailp` yields the following error:
``` bash
error: failed to get package in current working directory, pass `--workspace` if running from a workspace root
```
Adding `--workspace` as suggested results in a succsessful build of the `dailp` library alongside a "no queries found" warning. This message on its own is insufficient in furthering a root cause analysis of the initial error.

A series of google searches relating to issues in `sqlx prepare` eventually led to a GitHub discussion of a similar issue. A user in [this thread](https://github.com/launchbadge/sqlx/issues/788) suggested that issues of this nature could be a result of a mismatched versions of the `sqlx` crate and `sqlx-cli`. Looking at the version info in the dailp project, shown below, this seems to be the case.
```
sqlx-cli: 0.7.1
sqlx (crate): 0.6.3
```

## Solution
Updating the `sqlx` crate used by the `dailp` lib in `cargo.toml` fixed the issue, but required a number of configuration updates. First, sqlx 0.7 required upgrading channel in `rust-toolchain.toml` to use a more recent Rust version (previously using 1.72.0). Since Nixpkgs provides a maximum pinned Rust version, upgrading to the most stable Rust channel available seemed reasonable. The hash associated with `rust-toolchain.toml` in `flake.nix` also had to be updated. Second, crate dependencies in `flake.lock` needed to be updated via `cargo update`. Third, dereferences needed to be added in `database_sql.rs` when using functions relying on implementations of `sqlx:Execute` to accomodate [breaking changes in the implementation of sqlx's `Transaction` and `PoolConnection`](https://github.com/launchbadge/sqlx/blob/42ce24dab87aad98f041cafb35cf9a7d5b2b09a7/CHANGELOG.md#070---2023-06-30). Finally, additional Sqlx codegen checks and logging were added to `check.sh` to improve developer experience.


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.72.0"
+channel = "stable"
 profile = "minimal"
```
CharlieMcVicker, line 3: This scares me a little bit. Do we want this to bump without us explicitly changing anything?

CharlieMcVicker (commented):
Some questions/comments:

1. Does this part of our commit hook need to change now? Do we still have an `sqlx-data.json` or is the new `.sqlx` folder where this data lives now?
https://github.com/NEU-DSG/dailp-encoding/blob/main/.git-hooks/pre-commit#L9-L14
```shell
# Require sqlx data to be generated when a SQL query is modified.
if git-staged "types/queries/**.sql" && ! git-staged "sqlx-data.json"; then
    printf "\nDatabase queries changed without generating sqlx types.\nPlease run 'dev-check' or 'dev-generate-types'.\n"
    HOOKS_FAILED=1
fi
``` 
2. Why not add that `--workspace` flag and keep our sqlx query type data at the root level of the project like we have been? I think I like it living in `types` like it would now, but I wonder why we had done it at the root in the first place.
3. If `sqlx-data.json` is no longer where this data is actually stored, I'd like to see it deleted as part of the PR. I'm interested what will happen if you delete it and run this script.

I'm so exciting that you've found a fix for this and we can get moving on other stuff!

CharlieMcVicker (approved):
Just what we talked about! ᎣᏍᏛ ᎡᎵᏍᏗ 

ᏩᏙ.

PR #440, opened 2024-12-06
cm/safari audio hack

CharlieMcVicker:
- better indicate unsupported vs denied/unrequested media recorder status
- add extra logging to media recorder hook
- fix/hack: allow audio to playback on ios safari

Get around iOS safari never triggering the "oncanplaythrough" event, which our audio player waits for before showing the play button.


website/src/components/audio-player.tsx
```diff
@@ -3,6 +3,25 @@ import { FiLoader } from "react-icons/fi/index"
...
+/**
+ * Get our default load status for the audio player.
+ * On some user agents, we need to default to "loaded" because audio can never be loaded in the background.
+ */
+function getDefaultLoadStatus() {
+  // TODO: is there a place to store this instead of doing this check everytime?
+  // happens at this level because vite will run this server-side and crash sometimes
+  // @ts-ignore
+  const isIOS = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream
+  const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
```
GracefulLemming, line 15: It looks like audio loading is a problem across browsers on iOS (Tested on Chrome on an iPad using contributed audio from UAT). I know you said safari has some other particular strange behavior as well (mainly re: encoding). 

For load status, would it make sense to just check for iOS here to account for other browsers on iOS devices?

website/src/components/audio-player.tsx
```diff
@@ -3,6 +3,25 @@ import { FiLoader } from "react-icons/fi/index"
 import { MdPauseCircleOutline, MdPlayCircleOutline } from "react-icons/md/index"
 import * as css from "./audio-player.css"
 
+/**
+ * Get our default load status for the audio player.
+ * On some user agents, we need to default to "loaded" because audio can never be loaded in the background.
+ */
+function getDefaultLoadStatus() {
+  // TODO: is there a place to store this instead of doing this check everytime?
+  // happens at this level because vite will run this server-side and crash sometimes
```
GracefulLemming, line 12: My first thoughts are to do this through React state. I figure you considered that already though. Is there anything worse about using state than running this every time?

GracefulLemming (commented):
I left two comments, neither of them are blockers here but both are probably worth a short conversation.

website/src/components/audio-player.tsx
CharlieMcVicker: I believe all browsers on iOS have to send the safari user agent and are just safari reskins. There might be an exception with people who do "desktop view" which changes the user agent to not be safari. I don't know if that changes this behavior (not loading audio) though and I think it is a kind of a niche scenario.

website/src/components/audio-player.tsx
CharlieMcVicker: I'll look for a way to see if we are on the server-side, and short-circuit if we are? I wonder if that would let us make it a constant.

website/src/components/audio-player.tsx
GracefulLemming: Sounds good to me. Resolving.

GracefulLemming (approved):

PR #444, opened 2025-02-21
Phonetics Editing with warning for deleting phonetics

nole2701:
**For phonetics editing:**

- There is no phonetics before editing and there is no phonetics after editing -> do nothing
- There is no phonetics before editing and there is phonetics after editing -> save the new phonetics
- There is phonetics before editing and there is no phonetics after editing -> ask user if they mean to delete phonetics
- There is phonetics before editing and there is phonetics after editing -> if updated then save new, otherwise do nothing

pre-commit is commented out due to an issue with index.ts not updating after making a change to types in form.rs.


GracefulLemming:
@nole2701 When editing phonetics, GQL returns the error
``` JSON
{
    "data": null,
    "errors": [
        {
            "message": "Invalid value for argument \"word\", unknown field \"romanizedSource\" of type \"AnnotatedFormUpdate\"",
            "locations": [
                {
                    "line": 2,
                    "column": 14
                }
            ]
        }
    ]
} 
```

PR #452, opened 2025-04-18
Admin accounts 2 (backend)

### LINKED ISSUE #745: Get admin tools established

nole2701:
Back-end work done so far for admin accounts.


.git-hooks/pre-commit
```diff
@@ -1,40 +1,40 @@
...
-fi
-
-# Require formatting for rust code
-if git-staged "*.rs" && [[ -n "$(cargo fmt --check)" ]]; then
-    printf "Rust code edited but not formatted.\nPlease run 'dev-check'.\n"
-    HOOKS_FAILED=1
-fi
-
-exit $HOOKS_FAILED
+# #!/usr/bin/env bash
```
sarahkelley42: Removing this hook seems unrelated to the intent of the PR; can this be a separate PR?

admin-event-handlers/src/auth/post-confirmation/cognito_idp_operations.rs
```diff
@@ -33,4 +36,35 @@ impl CognitoClient {
...
+            .user_attributes(
+                AttributeType::builder()
+                    .name("username")
+                    .value(display_name)
+                    .build()?,
+            )
+            .message_action(MessageActionType::Suppress) // Suppress automatic welcome email
+            .send()
+            .await
+            .map_err(|e| anyhow::Error::new(e))?;
```
sarahkelley42, line 66: In the event this call fails, do we potentially end up with a stray admin user with a working temporary password, and if so do we have a mechanism for detecting and cleaning up such users?

graphql/src/query.rs
```diff
@@ -353,6 +357,15 @@ impl Query {
     async fn user_info<'a>(&self, context: &'a Context<'_>) -> Option<&'a UserInfo> {
         context.data_opt()
     }
+
+    /// Gets a dailp_user by their id
+    async fn dailp_user_by_id(&self, context: &Context<'_>, id: Uuid) -> FieldResult<User> {
```
sarahkelley42: What types of users have permissions to call this API, and where is that enforced?

graphql/src/query.rs
```diff
@@ -447,9 +460,7 @@ impl Mutation {
     }
 
     /// Mutation for adding/changing contributor attributions
-    #[graphql(
-        guard = "GroupGuard::new(UserGroup::Editors).or(GroupGuard::new(UserGroup::Contributors))"
-    )]
+    #[graphql(guard = "NotGroupGuard::new(UserGroup::Readers)")]
```
sarahkelley42: Does this change imply that a user who is not a member of _any_ group will have access to these APIs?

graphql/src/query.rs
```diff
@@ -536,6 +543,20 @@ impl Mutation {
             .await?)
     }
 
+    /// Updates a dailp_user's information
+    #[graphql(guard = "AuthGuard")]
+    async fn update_user(&self, context: &Context<'_>, user: UserUpdate) -> FieldResult<User> {
+        let user_id = Uuid::from(&user.id);
```
sarahkelley42, line 549: How does this API confirm that `user.id` is a user that the caller should have permission to update?

sarahkelley42 (commented):
Still working my way through, thanks for your patience on this. I think my big questions at the moment are who has permission to do what:

- Do we have mechanisms in place to prevent users with _no_ roles hitting the "not in group" auth guards? (deny by default is a good policy)
- For the insert user API, is this straight up just an insert into the database, or is there any workflow to get the user created in Cognito?
- For the update user API, do we have something that ensures that users are only able to update the correct set of users?
- What types of updates to a user are allowed, and by whom?

graphql/src/query.rs
nole2701: I allowed perms for all logged in users because that query is used for users editing their own profiles on their own profile page.

graphql/src/query.rs
nole2701: Since this is not referenced anywhere else in the codebase I believe I made that to use for manual testing purposes in the GraphQL Playground. I will comment it out for now.

graphql/src/query.rs
nole2701: Yeah I think I made a mistake there. There are some cases where a user can have no groups and call that mutation (as well as others that share the same negative guard). I'm changing all instances of this negative guard to a positive guard that allows specifically Editors/Contributors/Administrators..

admin-event-handlers/src/auth/post-confirmation/cognito_idp_operations.rs
nole2701: Two scenarios:
1. If the user creation fails, there is no user.
2. If the group assignment fails, there is a created user but **without any group**. It's a stray user.

I'm adding a rollback if the group assignment fails to clean up that stray user. If the rollback fails too, log the error.

.git-hooks/pre-commit
nole2701: At the time we had an issue with dev-check. I'll uncomment it now to match main.

admin-event-handlers/src/auth/post-confirmation/cognito_idp_operations.rs
sarahkelley42: Is this something that could reasonably be handled using a postgres transaction? If we can lean on commit/abort then it achieves the same result, without having an intermediate state where a user is potentially visible in the database

graphql/src/query.rs
sarahkelley42: To clarify, it looks like the request sends the user ID as part of the graphql query, and that's the ID that's used. It's possible for a logged-in user to supply a user ID that is not their own; is there anything on the backend that will check that the user being modified is the authenticated user?

admin-event-handlers/src/auth/post-confirmation/cognito_idp_operations.rs
nole2701: This function does not touch postgres so a postgres transaction cannot be used. It's purely a Cognito operation. The cognito calls `admin_create_user` and `add_user_to_group` are independent and are handled sequentially. As far as I know there's no way to group them in a transaction-equivalent.

graphql/src/query.rs
nole2701: That makes sense. I'll make an edit to throw an error if the requestor's id does not match the target user id. With an exception being made for `Administrators`

PR #454, opened 2025-04-28
Dw admin tool dasboard

denniwang:
Added support for admin tools frontend. Added admin role enum to user roles (not implemented in the backend yet)
Todo: 
- Adding admin role in AWS ( right now we dont have any admin accounts in the database )
- adding actual urls in the links

image


GracefulLemming (approved):
this all looks good, just make sure `.git-hooks/pre-commit` and `check.sh` are replaced with their counterparts on main as you merge main back in to this branch.

PR #467, opened 2025-06-13
Fix-website-type-errors

GracefulLemming:
Resolved a number of emergent type errors encountered in dev-check. 

1. replaced `Dailp.comment` with `Dailp.CommentFieldsFragment` in calls to `CommentBody` and `CommentHeader`

2. rewrote the existing code in authExchange as is (meaning no change made). I have no clue why this broke the way it did or why this fix worked.


nole2701 (approved):
Looks good to me.

PR #474, opened 2025-07-10
Improved error logging for Sheet Interpretation in spreadsheets.rs

nole2701:
- Made some changes on top of the previous commit. Some validation has been removed or made to throw warnings due to sheets sometimes having bad values.
- Improved error messaging so that it is easier to tell where errors are coming from, like what function and what sheet URL. 
- Commented out the part that fetches references because they're broken right now. Mentioned in console when it skips over it during migration.


GracefulLemming:
Included in #477

GracefulLemming:
Included in #477

PR #477, opened 2025-07-23
Improving spreadsheet.rs error logging 2

nole2701:
1. Improved error logging for all other functions in spreadsheets.rs outside SheetInterpretation.
2. Added error logging to SheetResult to give better details when failing to fetch sheets.
3. Added a log for which Google Sheet is failing when processing nouns/adjectives in lexical.rs


GracefulLemming (changes requested):
Please address all comments. 

It looks like generally, functions in spreadsheets.rs now follow the structure validate > process > log. We probaby want to abstract this structure, but we can do this later

PR #479, opened 2025-08-02
Added new metadata types

alisony755:
- Added Creator, Keyword, SubjectHeading, Format, Language, and SpatialCoverage types
- Added a Status enum to represent the status (pending, approved, rejected) of a suggested keyword, subject heading,etc.
- All instances of these types could have a status (whether they were suggested or not)
    - This could allow us to filter the keywords table by status to see all keywords that are approved as part of ourcontrolled vocabulary
- Added a ContributorRole enum (transcriber, translator, editor, etc.) to be able to assign more defined roles
- Added a description attribute to DocumentCollection type to describe a collection's contents, significance, etc.
- Edited Contributor type:
    - Revised just the description of the Contributor name attribute to be their name or preferred identifier (or“Anonymous” if no name or identifier provided) instead of their full name in case contributors want some level ofprivacy
    - Changed role attribute from Scalars[“String”] to ReadonlyArray to assign more defined roles
- Edited ContributorDetails type:
    - Added isPublic attribute to control whether or not the contributor's profile is linked to their contributions
    - Drafted logic (CitationFormatToFormats and FormatToCitationFormat) to map specific formats to their broader citationformat
        - Ex. A website and a blog post are distinct formats that share similar citation fields, so both would be mapped to amore general website citation format

TODO:
- Add TK (traditional knowledge) labels
- Add more fields to Citation type to cover a variety of formats
- Move CitationFormatToFormats and FormatToCitationFormat to appropriate section or file
- Add more mappings to CitationFormatToFormats as more formats are added to our database
- Also need to revise current CitationFormatToFormats mappings, as they may be inaccurate for some cases (ex. aYouTube video and film may not have the same citation format if the film is not online)
- Revise ContributorRole to accurate capture the roles of contributors (could also add uploader as a role?)
- Define dailpTags for Language and SpatialCoverage

### DIFF
--- a/website/src/graphql/dailp/index.ts
+++ b/website/src/graphql/dailp/index.ts
@@ -368,10 +368,14 @@ export type ContentBlock = Gallery | Markdown
 export type Contributor = {
   readonly __typename?: "Contributor"
   readonly details: Maybe<ContributorDetails>
-  /** Full name of the contributor */
+  /** 
+   * Name or preferred indentifier (ex. initials) of the contributor 
+   * Use "Anonymous" if no name or identifier provided or if contributor
+   * does not wish to be identified.
+   */
   readonly name: Scalars["String"]
   /** The role that defines most of their contributions to the associated item */
-  readonly role: Scalars["String"]
+  readonly role: ReadonlyArray<ContributorRole>
 }
 
 /**
@@ -397,6 +401,8 @@ export type ContributorDetails = {
    * them elsewhere, like in the attribution for a particular document.
    */
   readonly fullName: Scalars["String"]
+  /** Whether or not the contributor's profile is linked to their contributions */ 
+  readonly isPublic: Scalars["Boolean"] 
 }
 
 /**
@@ -492,6 +498,8 @@ export enum DocCitationFormat =
 
 export type DocumentCollection = {
   readonly __typename?: "DocumentCollection"
+  /** Description of the collection's contents, significance, etc. */
+  readonly description: Maybe<Scalars["String"]>
   /**
    * All documents that are part of this collection
    * TODO Try to unify this return type into AnnotatedDoc
--- a/website/src/graphql/dailp/index.ts
+++ b/website/src/graphql/dailp/index.ts
@@ -496,6 +496,38 @@ export enum DocCitationFormat =
   Video = "Video" // YouTubeVideo, Film
   // add more
 
+/** 
+ *  Draft of citation format mappings that can be relocated to a more 
+ *  appropriate file in the future.
+ */
+
+/** 
+ *  Maps broader citation format to corresponding formats. 
+ *  Many different formats may share the same citation format, so mapping 
+ *  broader citation formats to their specific formats may allow us
+ *  to automatically display specific citation input fields when a
+ *  contributor is entering the citation details for a document.
+ */
+export const CitationFormatToFormats: Record<DocCitationFormat, string[]> = {
+  [DocCitationFormat.Book]: ["Book", "EBook"],
+  [DocCitationFormat.Journal]: ["JournalArticle", "Newsletter"],
+  [DocCitationFormat.Website]: ["Website", "BlogPost", "Database"],
+  [DocCitationFormat.Audio]: ["Podcast", "RadioClip", "OralHistory"],
+  [DocCitationFormat.Video]: ["YouTubeVideo", "Film"],
+  // add more as needed
+}
+
+/** 
+ *  Maps specific formats to their broader citation format.
+ *  May be helpful to automatically display specific citation input fields 
+ *  when a contributor is entering the citation details for a document.  
+ */
+export const FormatToCitationFormat: Record<string, DocCitationFormat> = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


website/src/graphql/dailp/index.ts
alisony755, line 1: Need to add TK (traditional knowledge) labels

website/src/graphql/dailp/index.ts
```diff
@@ -372,6 +401,36 @@ export type ContributorDetails = {
...
   readonly fullName: Scalars["String"]
+  /** Whether or not the contributor's profile is linked to their contributions */ 
+  readonly isPublic: Scalars["Boolean"] 
+}
+
+/**
+ * A contributor can have to any number of roles, which define most of their
+ * contributions to the associated item
+ */
+export enum ContributorRole {
```
alisony755, line 412: May not need this, might be inflexible and hard to define new roles

website/src/graphql/dailp/index.ts
```diff
@@ -250,6 +250,31 @@ export enum CherokeeOrthography {
...
+  readonly publicationYear Scalars["Int"]
+  /** Publisher of the document */
+  readonly publisher Scalars["String"]
+  /** Starting page of the document */
+  readonly startPage Scalars["Int"]
+  /** Title of the document being cited */
+  readonly title: Scalars["String"]
+  /** URL of the document, if document can be accessed online */
+  readonly url Scalars["String"]
+  // TODO: add more fields to cover more formats
```
alisony755, line 274: Need to add more fields to cover citations for all formats

website/src/graphql/dailp/index.ts
```diff
@@ -577,6 +696,38 @@ export type ImageSource = {
...
+
+/** Record to store a language associated with a document */
+export type Language = {
+  readonly __typename?: "Language"
+  /** 
+   *  Tag for the language within the DAILP system 
+   *  Could be useful for managing similar language names or extending this to
+   *  add tags for language, dialect, and script combinations later on
+   */
+  readonly dailpTag: Scalars["String"]
```
alisony755, line 720: Define DAILP tags for language and spatial coverage

website/src/graphql/dailp/index.ts
```diff
@@ -425,8 +484,55 @@ export type DeleteContributorAttribution = {
   readonly documentId: Scalars["UUID"]
 }
 
+/** 
+ *  Represents the format of a citation. 
+ *  Used to map specific format to their more general citation format.
+ */
+export enum DocCitationFormat =
```
nole2701, line 491: I would recommend changing this enum to all caps since the rest of the enums in this file look like that. So like 
`Website = "WEBSITE"`

website/src/pages/documents/document-title-header.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <img
+            src={doc.thumbnailUrl ?? "https://www.wavonline.com/a/img/no_image_available.jpeg"}
+            alt={`Thumbnail for ${doc.title}`}
+            className={css.thumbnail}
+          />
+        )}
+
+        <div className={css.headerText}>
+          {breadcrumbs?.length ? (
+            <Breadcrumbs aria-label="Breadcrumbs" className={css.breadcrumbWrapper}>
```
nole2701, line 63: Navigation breadcrumbs looks clean.

nole2701 (commented):
Apart from changing the enum values to all caps in DocCitationFormat, I don't really see any jarring issues.

types/src/person.rs
```diff
@@ -10,7 +10,7 @@ pub struct Contributor {
     /// Full name of the contributor
     pub name: String,
     /// The role that defines most of their contributions to the associated item
-    pub role: String,
+    pub role: ContributorRole,
```
nole2701, line 13: This change here will require some refactoring across the back-end. There are some parts in the codebase that expect the role as a string. So we would definitely want a function that can swap between string and ContributorRole.

For example spreadsheets.rs line 870 it builds a contributor with the role parsed as a string from the spreadsheet it was taken from.

 // Process contributors with validation
        let people = names
            .into_iter()
            .skip(1)
            .zip(roles.into_iter().skip(1))
            .filter(|(name, role)| !name.trim().is_empty() || !role.trim().is_empty())
            .map(|(name, role)| Contributor { name, role })
            .collect();

image

This is what the spreadsheet looks like. Since the capitalisation is not consistent I'd recommend doing `to_lowercase()` then parsing into the appropriate role.

There's also 1 more spot in database_sql.rs at line 2010 where it creates a contributor from the SQL data, expecting a string. That's all that showed up when clicking "find references" for me but I'd double check on your branch just incase some changes were made.

Also, in the spreadsheet picture you can see Willie Jumper is a "Author" role which isn't part of the ContributorRole enum. I'd probably follow up about that.

nole2701 (commented):
Overall it looks pretty good. I would re-arrange the id field in these new objects to fit with the rest of the code, and address the refactoring issue that would arise from changing the type of role from string to enum.

PR #482, opened 2025-08-19
Create Collection at /collections/new

denniwang:
Added a description and thumbnail attribute to edited collections. Created a new route at `/collections/new` to let editors + admins create a new edited collection in the database, including a title, description, and a thumbnail image.

Home page with create document button (only visible to editors + admins)
image
Create document page
Screenshot 2025-08-19 at 10 32 47 AM

Todo:
- add migrations for cwkw and willie jumper in the db


types/queries/insert_edited_collection.sql
denniwang, line 1: renamed previous insert_collection.sql to insert_edited_collection, not sure if we want to allow users to add wordpress_menu_id as well.

PR #483, opened 2025-08-20
Dw create collection backend

denniwang:
- Added description + thumbnail url for edited collections in backend
- no frontend changes


nole2701 (commented):
Looks good. I would just clean up the migration files, and remove the transaction tag for the `insert_edited_collection` function.

graphql/schema.graphql
```diff
@@ -452,6 +452,24 @@ type ContributorDetails {
...
+"""
+input CreateEditedCollectionInput {
+	"""
+	The ID of the document to update
+	"""
+	title: String!
+	"""
+	An updated title for this document, or nothing (if title is unchanged)
+	"""
+	description: String!
```
alisony755, line 466: I think there might be some mismatches between the field names and their descriptions here:
- title is documented as "The ID of the document to update", which feels like it should be an id field instead
- description is documented as "An updated title...", which seems more like it belongs with title

alisony755 (commented):
Overall looks pretty good

types/migrations/20250812131535_add-description-to-edited-collections.sql
GracefulLemming, line 1: in the future, try to group migrations for a feature into one file.

types/src/raw/collection.rs
```diff
@@ -7,12 +7,16 @@ use uuid::Uuid;
 pub struct EditedCollection {
     /// UUID for the collection
     pub title: String,
+    /// Description of the collection (optional)
+    pub description: Option<String>,
```
GracefulLemming, line 11: As we review with the team, we should ask whether this should be optional from a workflow perspective.

Some questions we will have to answer:
1. Does not having a description for a collection even make sense?
2. Is getting a collection structure up quickly more important than adding a description upfront?

types/src/raw/collection.rs
```diff
@@ -7,12 +7,16 @@ use uuid::Uuid;
...
     /// Full title of the collection
     pub slug: String,
     /// ID of WordPress menu for navigating the collection
     pub wordpress_menu_id: Option<i64>,
     /// Chapters in the collection
     pub chapters: Vec<CollectionChapter>,
+    /// URL of the cover image for the collection
+    pub thumbnail_url: Option<String>,
```
GracefulLemming, line 19: We will want to have a similar conversation  here to the conversation I mentioned for the description. Does it make sense from a workflow perspective to allow this to be empty?

website/src/pages/edited-collections/utils.ts
GracefulLemming, line 1: For #482 , you should abstract out some of the code shared between here and the user contributed audio `utils.ts` you based this on.

PR #484, opened 2025-08-21
create document backend

denniwang:
Backend changes to support document creation. Literally #480 but I removed all the frontend changes. This is so we can deploy the changes to the GQL server on AWS so we can deploy on Netlify after.


nole2701 (commented):
Looks good overall, I would just remove the println message and keep an eye out during the merging process due to the ContributorRole enum replacing the string in the Contributor type from Alison's PR.

PR #499, opened 2025-09-23
Finished mobile view debugging (except orange ticket issues)

nole2701:
Issue list is in this spreadsheet
https://docs.google.com/spreadsheets/d/1mxTOXmBQQTaa4usXe2uaPrO0-COXD1Gc9EcekPqwNqI/edit?gid=0#gid=0

### Main Problems
- Text flowing beside headers creating narrow, unreadable columns
- Images causing horizontal overflow and appearing blurry
- Tables extending beyond screen width
- Videos misaligned with overlapping elements
- Form inputs in exercises overflowing off-screen

### Solution
- Added wordpressContentFix class applied conditionally upon `!isDesktop` detection
- Used globalStyle to target children elements within the WordpressPage using responsive sizing, as well as `clear` and `float` to stack content more vertically rather than horizontally
- added image and video handling to `const parseOptions` in wordpress.tsx to improve readability with `-webkit-optimize-contrast`, click to enlarge images in case it's a really big table that you can't read properly, and responsive sizing on all platforms

### Problems/Limitations
- Added an exception for 5+ column tables to allow them to overflow horizontally. It's not possible to squish them while keeping them readable so this is a compromise. I decided on 5+ columns from just looking around those pages and seeing what might and might not need it. A better fix would likely be in the Wordpress page design by perhaps rotating table elements, eg. 2x6 -> 6x2
- For the orange ticket issues, the redirect for all of them on the prev/next module buttons is either to a `dev.dailp.northeastern.edu/lessons/(page name)` or a wordpress link like `https://https//wp.dailp.northeastern.edu/lessons/(page name)`. I don't think the fix is in the codebase but rather in the Wordpress, just changing the redirects to `/collections/cwkw/(page name)` instead.

### DIFF
--- a/website/src/pages/edited-collections/chapter.css.ts
+++ b/website/src/pages/edited-collections/chapter.css.ts
@@ -101,5 +101,5 @@ globalStyle(`${wordpressContentFix} pre`, {
   maxWidth: "100%",
   margin: "0 0 1rem 0",
   padding: "0.5rem",
-  fontSize: "0.85rem"
+  fontSize: "0.85rem",
 })


website/src/pages/edited-collections/chapter.css.ts
```diff
@@ -13,3 +13,84 @@ export const docHeader = style([
...
+  `${wordpressContentFix} table td, ${wordpressContentFix} table th`,
+  {
+    padding: "0.4rem 0.2rem",
+    wordWrap: "break-word",
+    overflow: "hidden",
+    verticalAlign: "top",
+    width: "auto",
+    minWidth: "0px",
+    maxWidth: "200px",
+  }
```
alisony755, line 70: You could change the fixed maxWidth to a variable or theme token so it’s easier to tweak later

website/src/pages/edited-collections/chapter.css.ts
```diff
@@ -13,3 +13,84 @@ export const docHeader = style([
...
+globalStyle(`${wordpressContentFix} table td`, {
+  fontSize: "0.85rem",
+  lineHeight: "1.2",
+})
+
+globalStyle(`${wordpressContentFix} table td img`, {
+  width: "auto !important", // Override inline width="300"
+  height: "auto !important", // Override inline height="300"
+  minWidth: "120px !important",
+  display: "inline-block !important",
```
alisony755, line 87: The 120px minWidth could cause layout issues on small screens. Have we tested this on smaller devices?

website/src/components/wordpress.tsx
```diff
@@ -106,6 +106,57 @@ const parseOptions: HTMLReactParserOptions = {
...
+              maxWidth: "100%",
+              height: "auto",
+              imageRendering: "-webkit-optimize-contrast",
+              cursor: "pointer",
+              width: "auto",
+            }}
+            onClick={(e) => {
+              // Open image in new tab for full view
+              window.open(e.currentTarget.src, "_blank")
+            }}
```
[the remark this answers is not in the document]
nole2701: I just added a `title` tag saying "click to view full size". I thought about adding a caption that only shows on mobile but I don't think that's possible from the code end because the page comes from wordpress.

website/src/pages/edited-collections/chapter.css.ts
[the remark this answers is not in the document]
nole2701: I removed some stylings that are handled at the container level like for tables, but I left the width and maxwidth styles to maintain the sizing more explicitly incase on wordpress it might send in something different.

website/src/pages/edited-collections/chapter.css.ts
nole2701: I changed this to half the width of the screen.

PR #501, opened 2025-09-26
Dw/feature/admin menu editing backend

denniwang:
Same thing as PR #500, but I reverted all the frontend changes.


graphql/schema.graphql
```diff
@@ -800,6 +783,82 @@ type Markdown {
...
+	"""
+	id: UUID!
+	"""
+	Name for the menu.
+	"""
+	name: String!
+	"""
+	Slug for the menu.
+	"""
+	slug: String!
```
GracefulLemming, line 801: NTS: what is slug used for

types/src/database_sql.rs
```diff
@@ -1631,6 +1623,45 @@ impl Database {
                 .await?,
         )
     }
+
+    pub async fn get_menu_by_slug(&self, slug: String) -> Result<Menu> {
+        let menu = query_file!("queries/menu_by_slug.sql", slug)
+            .fetch_one(&self.client)
+            .await?;
+
+        let m_id: Uuid = menu.id;
```
GracefulLemming, line 1632: all three of these `m_` fields seem needlessly redundant since you're only using these values in 1639-41. Is there a reason you did this rather than just 
```
Ok(Menu {
        id: menu.id,
        name: menu.name,
        slug: menu.slug,
        items,
})

File: types/migrations/20250923142543_add-menu.sql [hunk 3]
```diff
@@ -0,0 +1,45 @@
+-- Menus
+create table menu (
+    id autouuid primary key,
+    name text not null,
+    slug text not null unique,
+    items jsonb not null
+);
+
+-- Seed default navigation menu (JSONB)
+insert into menu (id, name, slug, items) values (
```
GracefulLemming, line 10: if possible, it would be preferable to put _data_ migration code in our rust migration module

types/src/menu.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// Menu object representing the navbar menu that can be edited.
+#[derive(Debug, Clone, Serialize, Deserialize, async_graphql::SimpleObject)]
+pub struct Menu {
+    /// Id for menu.
+    pub id: Uuid,
+    /// Name for the menu.
+    pub name: String,
+    /// Slug for the menu.
+    pub slug: String,
```
GracefulLemming, line 18: replace slug with path for now

denniwang: the slug for the menu object is just the slugified version of the name, and it doesn't have anything to do with routing. it just makes it easier to find different menus.

PR #502, opened 2025-10-06
Page editing backend

denniwang:
- Lets users create new pages using markdown
- upsert pages, find page by path query
- page migration for existing graphql pages, importing from csv
- page contents use dailp gql instead of wordpress


migration/src/pages.rs
```diff
@@ -0,0 +1,50 @@
...
+        };
+        //println!("result: {:?}", page);
+        pages.push(page);
+    }
+    Ok(pages)
+}
+
+pub async fn migrate_pages(db: &Database) -> anyhow::Result<()> {
+    //println!("Migrating pages...");
+    let pages = load_pages("pages.csv")?;
```
GracefulLemming: I am getting a "no such file or directory" error here

migration/src/pages.rs
```diff
@@ -0,0 +1,50 @@
...
+    title: String,
+    content: String,
+}
+
+pub fn load_pages(file_path: &str) -> Result<Vec<NewPageInput>, anyhow::Error> {
+    let file = File::open(file_path)?;
+    let mut reader = ReaderBuilder::new().from_reader(file);
+
+    let mut pages = Vec::new();
+    for result in reader.deserialize() {
```
GracefulLemming: You should add some explicit error handling for CSV parse issues. What happens if a column is missing? if it is malformed?

GracefulLemming (changes requested):

PR #513, opened 2025-10-31
Add genre metadata

alisony755:
- Moved changes related to genre metadata into separate PR


denniwang:
I think you should run `dev-check` from the root dir, it'll format the code for you and do a bunch of other helpful checks.

PR #515, opened 2025-10-31
Add creator metadata

alisony755:
- Add creator metadata changes to separate PR

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -78,10 +78,6 @@ type AnnotatedDoc {
 	"""
 	chapters: [CollectionChapter!]
 	"""
-	Creators of this document
-	"""
-	creators: [Creator!]!
-	"""
 	Key terms associated with a document
 	"""
 	keywords: [Keyword!]!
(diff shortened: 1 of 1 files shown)


types/queries/get_document_creators.sql
GracefulLemming: Is this equivalent to `get_creators_by_document_id`

types/queries/many_creators_for_documents.sql
GracefulLemming: Not sure this will be used, consider removing

GracefulLemming (changes requested):
Please also delete redundant schema.graphql

types/src/database_sql.rs
```diff
@@ -770,16 +782,81 @@ impl Database {
...
             "queries/update_document_metadata.sql",
             document.id,
             &title as _,
             &written_at as _
         )
-        .execute(&self.client)
+        .execute(&mut *tx)
         .await?;
 
+        // Update creators
```
alisony755: Still tweaking

PR #516, opened 2025-10-31
Add keyword metadata

alisony755:
- Add keyword metadata changes to separate PR


nole2701 (commented):
Looks good, but I'm missing information on how the keywords table was initially created because I didn't see the PR where it was initially added. Since this PR makes approval_status not null, is everything else in data migration already set up to satisfy these null value constraints?

PR #518, opened 2025-11-01
Add subject heading metadata

alisony755:
- Move changes related to subject headings metadata field to separate PR


types/queries/many_subject_headings_for_documents.sql
GracefulLemming: It's unclear to me why we would need this. To load a document's headings we use `get_document_subject_headings`. To find documents by subject heading (which we don't even have scoped as a need right now...) we would use `get_documents_by_subject_headings`. What was your thinking here?

types/queries/get_document_subject_headings.sql
GracefulLemming: Conceptually, how is this different from `get_subject_headings_by_document`?

types/queries/many_subject_headings_for_documents.sql
GracefulLemming: please do!

types/queries/get_document_subject_headings.sql
GracefulLemming: Since our sql queries return info directly to our rust middleware, I don't see any advantage to maintaining both of these solutions. Pick one to keep please.

graphql/schema.graphql
```diff
@@ -81,6 +81,15 @@ type AnnotatedDoc {
...
 	"""
 	chapters: [CollectionChapter!]
+	"""
+	Subject heading IDs associated with this document
+	Seems overcomplicated?
+	"""
+	subjectHeadingsIds: [UUID!]!
+	"""
+	Internal field accessor for subject headings
+	"""
```
alisony755: Will try skipping this

PR #526, opened 2025-11-07
On-Premise User Accounts 

nole2701:
Finished local user accounts up until re-doing uploads. User experience should be exactly the same as with cognito. I will post the new .env file on Slack.

Currently working:
- Logins
- Signups
- Password Resets

Not working:
- Uploads as current S3 upload method is directly tied to cognito.

As for testing the tokens for user actions like editing comments, there's currently a migration problem with pages which prevents me from doing that. But after logging in, the AccountMenu component properly displays the email of the user. This means the access token successfully passed through the stack following the same path as cognito tokens and returned UserInfo.

Also, the dailp.noreply@gmail.com email is tied by 2FA to my phone (required for App password) so we would probably want to find a way to have it tied to our organisation.


website/src/auth.tsx
```diff
@@ -31,261 +40,134 @@ const userPool = new CognitoUserPool({
...
-export function getCurrentUser(): CognitoUser | null {
-  return userPool.getCurrentUser()
+export function getCurrentUser(): AuthUser | null {
+  if (AUTH_MODE === "cognito") {
+    const cognitoUser = userPool.getCurrentUser()
+    return cognitoUser ? { type: "cognito", user: cognitoUser } : null
+  } else {
+    const token = LocalStorage.getItem("dailp_access_token")
+    if (!token) return null
+    const decodedToken = jwtDecode<{
```
alisony755: `jwtDecode` could throw an error if the token is malformed or expired. You could consider wrapping it in a try/catch to avoid crashing

website/src/auth.tsx
```diff
@@ -31,261 +40,134 @@ const userPool = new CognitoUserPool({
...
+      sub: string
+      email: string
+      role: string
+      exp: number
```
alisony755: You could consider checking `exp` against the current time and return null if the token is already expired

graphql/src/query.rs
```diff
@@ -889,6 +893,164 @@ impl Mutation {
...
+        let db = context.data::<DataLoader<Database>>()?.loader();
+
+        // Create user account
+        let user_id = db.create_user(input.email.clone(), input.password).await?;
+
+        // Generate verification code
+        let verification_code = db.create_email_verification_token(user_id).await?;
+
+        // Send verification email
+        email::send_verification_email(&input.email, &verification_code).await?;
```
alisony755: It seems like signup would fail if the verification email doesn't send successfully? You could consider adding a retry/resend option

website/src/auth.tsx
nole2701: This shouldn't be needed because of the setupTokenRefresh() function handling expired tokens, and the back end will block expired tokens too. I ended up removing `exp` entirely because I didnt end up using it anyways.

alisony755 (commented):
I tested it and it works with the updated .env file!

doc/database/user.md
```diff
@@ -2,19 +2,74 @@
...
+| `display_name`   | `text`      | How the user's name should be presented in the app                         |
+| `created_at`     | `date`      | When the user record was created                                           |
+| `avatar_url`     | `text?`     | Profile picture URL                                                        |
+| `bio`            | `text?`     | User biography                                                             |
+| `organization`   | `text?`     | User's organization                                                        |
+| `location`       | `text?`     | User's location                                                            |
+| `role`           | `user_role` | User's role, defaults to `Readers`                                         |
+| `email`          | `text?`     | Email address (local accounts only, unique)                                |
+| `password_hash`  | `text?`     | Argon2 password hash (local accounts only)                                 |
+| `email_verified` | `boolean?`  | Whether email has been verified (local accounts only), defaults to `false` |
```
sarahkelley42: If it defaults to `false`, can this field be made non-nullable?

doc/database/user.md
```diff
@@ -2,19 +2,74 @@
...
+| `id`             | `uuid`      | Primary key                                                                |
+| `display_name`   | `text`      | How the user's name should be presented in the app                         |
+| `created_at`     | `date`      | When the user record was created                                           |
+| `avatar_url`     | `text?`     | Profile picture URL                                                        |
+| `bio`            | `text?`     | User biography                                                             |
+| `organization`   | `text?`     | User's organization                                                        |
+| `location`       | `text?`     | User's location                                                            |
+| `role`           | `user_role` | User's role, defaults to `Readers`                                         |
+| `email`          | `text?`     | Email address (local accounts only, unique)                                |
+| `password_hash`  | `text?`     | Argon2 password hash (local accounts only)                                 |
```
sarahkelley42, line 18: I'm rusty on password storage security. Does Argon2 eliminate the need for a salt? Thinking about whether reused passwords (or multiple accounts that happen to have the same password) would have the same hash.

Cargo.lock
```diff
@@ -141,6 +141,27 @@ version = "1.0.94"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
 
+[[package]]
```
sarahkelley42, line 144: Shouldn't we also update Cargo.toml? General .lock is machine-generated (used for pinning exact versions) while .toml is the human-facing file for specifying intent.

I see that Cargo.toml is pretty sparse, not sure if this was intentional or not for DAILP. I'd generally recommend updating that, so that we have a clear idea of what the direct dependencies are (as opposed to the transitive dependencies which may over time need to get pruned away)

graphql/src/dailp_auth.rs
```diff
@@ -0,0 +1,22 @@
+use dailp::auth::{DailpJWT, UserInfo};
+use jsonwebtoken::{decode, Algorithm, DecodingKey, Validation};
+
+const BEARER: &str = "Bearer ";
+
+pub fn user_info_from_authorization(auth: &str) -> Result<UserInfo, anyhow::Error> {
```
sarahkelley42, line 11: Question: What happens to sessions if the JWT key needs to change? Assuming this gets rotated, or at least that we'll want some mechanism for changing it if it leaks. Do we need to do anything special to revoke existing sessions, or do they simply start getting rejected as we roll out servers with the new config? What could a random person do if they had this key?

Couple suggestions on this bit:
- Since this function ends up being an important piece of the app's security, I think it'd be useful to have this piece unit tested using Rust's #[test] feature.
- That said, unit tests are going to be tricky if we're reading environment variables directly; do we have any concept of dependency/config injection so we can fake it in the test?

graphql/src/query.rs
```diff
@@ -968,6 +972,237 @@ impl Mutation {
...
+
+        Ok(MessageResponse {
+            message: "Account created successfully! Please check your email for your 6-digit verification code.".to_string(),
```
sarahkelley42: Assuming this message is displayed to the user: what does the user see if we hit the error cases above?

graphql/src/query.rs
```diff
@@ -968,6 +972,237 @@ impl Mutation {
...
+            return Err("Local user accounts are not enabled".into());
+        }
+
+        let db = context.data::<DataLoader<Database>>()?.loader();
+
+        // Check if user exists (don't reveal if they don't for security)
+        if let Some(user) = db.get_user_by_email(&input.email).await? {
+            // Create and send password reset token
+            let reset_token = db.create_password_reset_token(user.id).await?;
+            email::send_password_reset_email(&input.email, &reset_token).await?;
```
sarahkelley42: In each of these cases, what happens if we hit the error case of `.await?` ? Does that leak any info to the user?

graphql/src/query.rs
```diff
@@ -968,6 +972,237 @@ impl Mutation {
...
+        // Always return success to prevent email enumeration
+        Ok(MessageResponse {
+            message: "If an account with that email exists, a password reset link has been sent."
+                .to_string(),
+        })
+    }
+
+    /// Reset password using reset token (DAILP auth only)
+    async fn reset_password(
+        &self,
```
sarahkelley42, line 1115: I'm noting that a lot of the methods being added in this PR don't use `self` at all. This starts to suggest to me that these could be a separate module for user account management, with these functions being free functions not attached to any particular object. This has the nicety that we wouldn't have to track ownership of the `Mutation`, and that we could run these functions independently in tests without needing to construct a `Mutation`. Design-wise, is there a particular need for these concepts to be attached to a `Mutation` directly?

graphql/src/query.rs
```diff
@@ -968,6 +972,237 @@ impl Mutation {
...
+            return Err("Email verification is only available in DAILP authentication mode".into());
+        }
+
+        let db = context.data::<DataLoader<Database>>()?.loader();
+
+        // Get user by email
+        let user = db
+            .get_user_by_email(&email)
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("No account found with that email address"))?;
```
sarahkelley42: I see this method has responses that indicate the existence/nonexistence of an account. Could a malicious user hit this API to probe for logins?

sarahkelley42 (commented):
I'm focusing on the backend pieces for now as that's more my area of expertise. I'm still onboarding to the codebase, so take my suggestions with a grain of salt.

I'm making my way through the PR in pieces - consider this a work-in-progress review. :)

chullings:
@GracefulLemming The other of Sarah's reviews

doc/database/user.md
nole2701: Yes this makes sense. I updated it to be non-nullable.

doc/database/user.md
nole2701: Hashing the same password with Argon2 will generate the same hash so salting is needed to make each stored password unique.

Cargo.lock
nole2701: Cargo.lock is auto generated from Cargo.toml. In this project, the Cargo.toml file in the root lists the workspace members being `["types", "graphql", "migration"]`. Those directories have their own Cargo.toml files which list all the dependencies each one uses.

graphql/src/dailp_auth.rs
nole2701: If the JWT Secret gets rotated, users will need to log in again after their current access token expires. Refresh tokens are not affected since they're not signed with the JWT secret. We don't really need to do anything special to revoke existing sessions, the sessions will just renew over the hour that old access tokens will expire.

If someone had this JWT secret, they could forge an access token (for any role including admin) and impersonate anyone on the website until the token expires.

For unit tests, I think injecting an env should be possible, using the `std::env::set_var` function.

graphql/src/query.rs
nole2701: Yes the user would see an error as its sent up the stack and displayed in an alert. I implemented a fix that made all the error messages generic, and contained them in the back end using log::error/log::info, as well as catching an error for create_user that returns Ok(None) if a duplicate sign-up were to happen.

graphql/src/query.rs
nole2701: Yes this old implementation would leak info if the SMTP failed for an existing email (non-existing email got a generic response). I implemented a fix by containing the errors to the backend and changing response messages to all be generic.

graphql/src/query.rs
nole2701: Yes it's a leak. I fixed it by returning only generic messages now.

graphql/src/query.rs
nole2701: For everything in query.rs right now, &self is required so that context be used as a parameter, which is how we reference the database.

For tests, I moved all the auth related logic from the mutation resolvers in query.rs to dalip_auth.rs as helper functions. This should make them easier to write tests for. The mutations remain but the logic moved.

graphql/src/dailp_auth.rs
```diff
@@ -0,0 +1,246 @@
...
+                user_id,
+                e
+            );
+            return Ok(generic_message);
+        }
+    };
+
+    if let Err(e) = email::send_verification_email(&input.email, &verification_code).await {
+        log::error!("signup: failed to send verification email: {}", e);
+    }
```
sarahkelley42, line 76: Non-blocking: I'm seeing that each of the error conditions still sends an OK response back to the user. If there's a legit error (e.g. it's a legit new user, but creating the account failed, or creating the account succeeded but sending the verification email failed), is there a way for either 1/ the user to know that the error occurred, or 2/ an admin to find out that there was an error and take appropriate corrective action?

I think this is the right default error handling structure if we're unsure; it avoids leaking information to a malicious user and it logs appropriately so that we can figure out what happened later. We can always improve error handling in a followup.

graphql/src/dailp_auth.rs
```diff
@@ -0,0 +1,246 @@
...
+
+fn require_dailp_mode(feature: &str) -> anyhow::Result<()> {
+    if std::env::var("AUTH_MODE").unwrap_or_else(|_| "cognito".to_string()) != "dailp" {
+        return Err(anyhow::anyhow!(
+            "{} is only available in DAILP authentication mode",
+            feature
+        ));
+    }
+    Ok(())
+}
```
sarahkelley42: Non-blocking: do we plan to use the `feature` string in future work? e.g. mapping modes to features. As-is it seems like this function is a fancy boolean around `AUTH_MODE == "dailp"` which is reasonable, but at the call sights it looks like the features may be inputs to the decision somehow.

migration/src/main.rs
```diff
@@ -59,7 +59,7 @@ async fn main() -> Result<()> {
     menu::migrate_menu(&db).await?;
 
     //println!("Migrating collections...");
-    //edited_collection::migrate_edited_collection(&db).await?;
+    edited_collection::migrate_edited_collection(&db).await?;
```
sarahkelley42: Blocking: Is including this change intentional? Not sure if this is related to on-prem user accounts. If it's intentional that's fine, I want to ask about this before considering this merge ready b/c it seems possibly mutating/destructive.

types/src/database_sql.rs
```diff
@@ -2706,6 +2718,441 @@ impl Database {
...
+
+        let password_hash = Self::hash_password(&password)?;
+        let normalized_email = email.to_lowercase().trim().to_string();
+
+        match query_file_scalar!("queries/create_user.sql", &normalized_email, password_hash)
+            .fetch_one(&self.client)
+            .await
+        {
+            Ok(user_id) => Ok(Some(user_id)),
+            Err(sqlx::Error::Database(e)) if e.code().as_deref() == Some("23505") => Ok(None),
```
sarahkelley42, line 2979: Non-blocking: what does 23505 correspond do?

doc/database/user.md
sarahkelley42: Ah, I see that it's salted with the Argon library itself. Thanks!

sarahkelley42 (commented):
Generally LGTM. I have one blocking question on the migration code that's now uncommented but I just need clarification, if it's intentional I think that's fine. I have a few non-blocking thoughts but I don't think they're necessary pre-merge.

I was focusing on the backend and database code for this review; I'm assuming y'all know more than me about React, and frontend changes are more easily changeable than database/schema type concerns.

Some thoughts for future work:

1. Cleanup for expired tokens - right now we handle expiration via the SELECT, but over time this is going to get big. It's going to take a long time before that's a problem so I think we've got buffer, but we'll want something to handle that before giving this to communities that may not update the software for a long time. (the tail on software update schedules is _looong_ :) )
2. Error handling simplification: I think we could reduce a lot of the problem space if we make our operations idempotent. For example, sending a verification email is basically always safe, so we might not need a concept of "re-send" beyond just triggering the normal send.
3. Error reconciliation paths: if a legit user isn't getting a verification email it may be worth considering what (if anything) we can do to make it clear that there's a problem, while still not exposing unnecessary information to malicious users.

types/src/database_sql.rs
nole2701: It's the `UNIQUE_VIOLATION` error code for postgres

migration/src/main.rs
nole2701: At the time there were some issues with migration being broken so we commented out some of the migration. As migrations still don't work and we're using UAT data dumps to get our data for local testing, I'm not sure what to do here either.

graphql/src/dailp_auth.rs
nole2701: Those other errors will land in the dev-graphql terminal so admins still have visibility over them. If the SMTP fails, we already have a re-send verification email as a backup but I guess it could be nice for users to know if the SMTP is not working. We can do that as a follow up if needed.

graphql/src/dailp_auth.rs
nole2701: It just helps us identify which function triggered the error. I'll rename it to `error_label` to make the purpose clearer.

migration/src/main.rs
sarahkelley42: Migrations being broken seems like a sufficiently important problem that it warrants its own task and PR. In general I view this kind of thing from the POV of prod hygiene; we want to minimize surprises when deploying to a production environment, and conversely we want to ensure that one-way destructive actions (e.g. migrations) are extraordinarily obvious.

If we merge this PR without un-commenting this line, will the migrations introduced in this PR block the migrations that are commented out here when they're re-enabled later?

nole2701: In practice, including this line wouldn't matter since migrations are still broken now, and we get our data from dump files for now. But I'll just match this file to main so that the same migration functions are active and maintains the current main branch status quo.

doc/database/user.md
```diff
@@ -2,19 +2,74 @@
 
 ## `dailp_user`
 
-Metadata assocated with a user. `dailp_user.id` on this table is equal to `sub` in
-AWS. Users are not to be confused with `contributor` entires, which are imported
-from Google Sheets.
+Stores user account data. Supports two authentication modes controlled by the `AUTH_MODE` environment variable: AWS Cognito (`cognito`) or local accounts (`dailp`). When using Cognito, `id` corresponds to the Cognito `sub` claim; when using local accounts, `id` is generated at signup. `email` and `password_hash` are nullable to maintain compatibility with the Cognito auth mode (Cognito-authenticated users have no local password and may not have an email stored in this table). `email_verified` is non-nullable and defaults to `false`; Cognito users are not subject to the local verification gate, so the value is meaningless for them but storing `false` is harmless.
```
GracefulLemming, line 5: Non-blocking: 
This description suggests we may want to break this table into two tables: one for dailp user _profiles_ and one for dailp user _identities_, where the identities table is only used for local accounts.

graphql/src/query.rs
GracefulLemming: This context method of passing database connection info does cause a lot of unnecessary references to `self`. It also makes organizing our queries substantially harder, and generally feels like it's probably an anti-pattern. @sarahkelley42 let's think about how to refactor this.

website/vite.config.ts.timestamp-1762386481580-ec89897efe7f4.mjs
GracefulLemming, line 1: blocking: is this file needed in version control?

package-lock.json
GracefulLemming, line 1: Blocking: was this intentional? We use yarn and its associated lock file. We likely should not have a `package-lock.json`

package.json
GracefulLemming, line 1: Blocking: remove this file

GracefulLemming (changes requested):

PR #657, opened 2026-02-09
Homepage redesign

alisony755:
- Redesigned homepage
- Need to fix links in nav bar
- Need to fix WordPress related errors


website/src/pages/wordpress.page.route.ts
GracefulLemming: This file should not exist, it looks like maybe you made a mistake when merging main?

PR #666, opened 2026-02-20
Collection chapter backend fix

### LINKED ISSUE #660: Page publishing for collections

nguyen-katie:
This PR updates the CollectionChapter struct in which it differentiates between a page and a document. I updated the schema and the frontend now shows the correct chapters in the table of contents that can be published under. However, there is still no backend method for collections publishing yet that works similar to MenuUpdate, so the feature for collections is unfinished until then. I added to the alert for when someone wishes to change the location of a page in a collections that says the publishing feature is still in development, but I am happy to comment out the code again to cover up the feature for collections entirely.

image


website/src/pages/dailp.page.tsx
```diff
@@ -165,7 +175,10 @@ export const DailpPageContents = (props: { path: string }) => {
   }
 
   const locationList = isInCollection
-    ? collectionSections?.map((item) => (
+    ? [
+        ...(collectionSections ?? []),
+        ...(collectionPageChapters?.map((ch) => ch.title) ?? []),
+      ].map((item) => (
         <option key={item} value={item}>
```
nole2701, line 182: Would there be a way to distinguish between sections and chapter titles that have the same name when in the menu dropdown?

nole2701 (commented):
Looks good just that one nitpick about duplicate names that might happen.

PR #694, opened 2026-03-31
Added free keyword text area

### LINKED ISSUE #652: Free text keywords
Edit the Keywords field so that it is free text and contributors + editors can type values

### LINKED ISSUE #670: Spatial Coverage Free Text Inputs
Add the ability for editors to input free text keywords that will then be added to the drop down

Please ensure free text limit needs to be the length of a single line in the drop-down (likely around 40 characters)

SonAndrea:
Allows for users to input a maximum of 40 characters for a keyword when editing meta data. Additionally, kept the selector for preexisting keywords for easy of use.


website/src/pages/documents/tag-selector.tsx
```diff
@@ -10,6 +10,7 @@ interface TagSelectorProps {
   onRemove?: (index: number) => void
   addButtonLabel: string
   customForm?: React.ReactNode
+  children?: React.ReactNode
```
alisony755: It might be helpful to be slightly more clear about the purpose of `children`, either with a comment or renaming it to something like `freeInput`.

alisony755 (dismissed):
This looks good and works well! I like the option to add either an existing keyword or a custom one.

website/src/pages/documents/tag-selector.tsx
alisony755: Or `additionalForm` like you use for subject headings

alisony755 (approved):
Sorry my previous comments were not accurate! I was not logged in with an editor account, which is why editing wasn't working. I tested it again and the free text entry for keywords works well!

website/src/pages/documents/edit-document-modal.tsx
```diff
@@ -792,8 +793,45 @@ export const EditDocumentModal: React.FC<EditDocumentModalProps> = ({
...
+                }}
+              >
+                <input
+                  type="text"
+                  className={styles.input}
+                  placeholder="Enter keyword..."
+                  value={freeKeyword}
+                  onChange={(e) => setFreeKeyword(e.target.value)}
+                  onKeyDown={(e) => {
+                    if (e.key === "Enter" && freeKeyword.trim()) {
```
GracefulLemming: I like this shortcut, but it does have overlap with the submit button which may be disorienting for users. You don't need to change this PR but I think the solution could involve the submit button only working on focus (assuming that solution doesn't run us astray of accessibility compliance).

@chullings and @jae-mess we should think more about keyboard use across the site

website/src/pages/documents/edit-document-modal.tsx
```diff
@@ -792,8 +793,45 @@ export const EditDocumentModal: React.FC<EditDocumentModalProps> = ({
             newTags={newKeywords}
             onAdd={(tagName) => addKeyword(tagName)}
             onRemove={removeKeyword}
-            addButtonLabel="Add Keyword"
-          />
+            addButtonLabel="Add Pre-existing Keywords"
+          >
+            {isEditing && (
```
GracefulLemming: @chullings are there any metadata fields that _shouldn't_ allow extension from editors? 

If we can tolerate all metadata fields drawing from user-built dictionaries like keywords does here, then I think we can move the responsibilities within this block into the `TagSelector` component

website/src/pages/documents/edit-document-modal.tsx
```diff
@@ -792,8 +793,45 @@ export const EditDocumentModal: React.FC<EditDocumentModalProps> = ({
...
+              <div
```
GracefulLemming: Do we need two input fields? 

"Add Keyword" and "Select Pre-existing Keyword" both interface with a dictionary of keywords (stored in the keyword table). Why not collapse these fields like we do with word parts? 

So, when a user types into the input field, they receive hints pulled from the dictionary (using a substring match). A user can add an existing keyword to a document's metadata by selecting a keyword from the hint list or typing out a keyword in full that is already present in the dictionary. A user can add a keyword that is not in the dictionary by typing it in full before pressing "Add Keyword" (or pressing enter in this case); In this case, the keyword would be added to both the document metadata and the dictionary.

GracefulLemming (commented):
No blocking comments here, just some food for thought.

website/src/pages/documents/edit-document-modal.tsx
SonAndrea: I removed the shortcut as while nice, I didn't see that conflict and think that it is best to leave it up to site-wide decisions (especially since the other fields don't have it).

website/src/pages/documents/edit-document-modal.tsx
SonAndrea: You're technically right. At the time of writing this code, I believe that contributors were allowed to edit metadata but only editors were allowed to add new keywords. This check is redundant now since only editors can write into metadata.

website/src/pages/documents/edit-document-modal.tsx
SonAndrea: As for this pr, I think this is fine by the specs. In the future I believe all fields will work as you've just described as I've done this with later work for the subject headings.

PR #707, opened 2026-04-07
Added Metadata Tooltip Information Modals

### LINKED ISSUE #671: Metadata Information Modals
Add "Information" icons for modals over the Metadata Types listed here: 

- Date Created
- Document Type
- Format
- Contributors
- Keywords
- Subject Headings
- Spatial Coverage

Descriptions to use in the doc here:
https://docs.google.com/document/d/1WLaeRl_njWYD6aS_9MwPfh1irEtR91PCmXmO3YeeLSY/edit?usp=sharing

SonAndrea:
Added information modal popups as according to task #671 Metadata Information Modals for metadata fields: 
- Date Created
- Document Type
- Format
- Contributors
- Keywords
- Subject Headings
- Spatial Coverage


website/src/components/info-tooltip.css.ts
```diff
@@ -0,0 +1,61 @@
+import { style } from "@vanilla-extract/css"
+import { colors } from "src/style/constants"
+
+export const container = style({
+  position: "relative",
+  display: "inline-flex",
+  alignItems: "center",
+  marginLeft: "6px",
```
nguyen-katie: looks great! Only thing I'd suggest is to avoid pixels/px units & using rem instead if possible according to the [code standards](https://github.com/NEU-DSG/dailp-encoding/wiki/Code-Standards), but no changes needed here since the tooltips still look nice in smaller screens.

website/src/components/info-tooltip.tsx
nguyen-katie, line 1: I would suggest not making a new tsx file for just info tooltips since it will probably be used for metadata only. It might be better to condense files by making `InfoTooltip `an export of  `edit-document-modal.tsx` and adding the associated styles to `edit-document-modal.css.ts`

website/src/pages/documents/tag-selector.tsx
nguyen-katie, line 1: I would suggest not making a new tsx file for just info tag selectors since it will probably be used for metadata only. It might be better to condense files by making `TagSelector`an export of  `edit-document-modal.tsx` and adding the associated styles to `edit-document-modal.css.ts`

SonAndrea:
Made changes to remove use of pixels in css for tooltips. I did not makes the changes of moving the tooltip login into the metadata as there are potential uses for the component outside of just using it within metadata tooltips.

alisony755 (changes requested):
This looks good, but the subject headings tooltip doesn't seem to have been added. I might also add a black border around the tooltip to help distinguish it from the background.

SonAndrea:
Thanks for catching subject heading being removed; occurred when merging main. Added the subject heading content and black border to tooltips.

alisony755 (approved):
Thanks for making those changes. This looks good!

PR #777, opened 2026-05-26
#754 Delayed Page Formatting

### LINKED ISSUE #754: Delayed Page Formatting
## Description of the Bug
**Expected behavior:** When a new page is published, formatting will automatically apply.
**Actual behavior:** When a new page is published, there is no formatting at all
## Symptoms of the problem
**Context:** Happens for any new pages created as an editor. Seen on UAT in a Firefox browser.
## Related Features
Page creation and page publishing
## Likely Problem and Solution
The formatting gets added once the site rebuilds. Formatting is only getting applied to pages that already exist from that build. We need to converge the pathways from existing content and newly published pages so formatting applies immediately.

nole2701:
Adding the same formatting wrapper around content in `_error.page.tsx` as `dailp.page.tsx` should cause this fallback to apply proper initial formatting. Verified using "Page Not Found" pages which pass through `_error.page.tsx`.

### DIFF
--- a/website/src/renderer/_error.page.tsx
+++ b/website/src/renderer/_error.page.tsx
@@ -5,6 +5,7 @@ import * as Dailp from "src/graphql/dailp"
 import Layout from "src/layout"
 import { DailpPageContents } from "src/pages/dailp.page"
 import { useLocation } from "src/renderer/PageShell"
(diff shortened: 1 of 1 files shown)


chullings:
@GracefulLemming Please make sure this will work for the collections (Create New Document) as well as pages

PR #784, opened 2026-06-04
added delete confirmation for comments

### LINKED ISSUE #735: Word Pane Editor Warnings
- "Discard” button on Comment did not have a warning message (“this will delete all content”). Add a warning message.

SonAndrea:
When a user now uploads a comment and deletes it, instead of automatically being deleted, a confirmation is displayed. I used the same styling of the popup for the audio deletion pr; however, in the future, all similar confirmation popups should be abstracted under a unified component (in other words, the current styling should be modified and abstracted).


website/src/components/comment-section.css.ts
```diff
@@ -109,3 +109,52 @@ export const deleteButton = style([
     width: "auto",
   },
 ])
+
+// Warning popup pulled from delete audio popup
+// * In the future consolodate popups under a comp or adjust this as needed *
```
GracefulLemming, line 114: Is there a reason you chose not to make a dedicated component for this?

website/src/components/comment-section.tsx
```diff
@@ -152,6 +153,30 @@ export const CommentBody = (p: { comment: Dailp.CommentFieldsFragment }) => {
...
+
+            <div className={css.modalButtonGroup}>
+              <Button onClick={() => setIsConfirmingDelete(false)}>
+                Cancel
+              </Button>
+
+              <Button
+                onClick={handleDelete}
+                style={{
+                  backgroundColor: "#b72d3b",
```
GracefulLemming, line 171: Is there a reason this is styled inline?

SonAndrea: I did it this way since I am using the global css for the button but wanted it red to indicate cancelation. This was a shortcut to implementing it as its own exported style.

website/src/components/comment-section.css.ts
SonAndrea: I wanted ux feedback before adding new files for a component in the event that the whole piece is not sufficient.

jae-mess:
I don't think we typically have red buttons, compare the Save/Cancel and Comment/Discard pairs using white and blue. I think confirm should be blue and cancel should be white following this style - let me know what you think. 
Screenshot 2026-09-17 at 10 23 29 AM
Screenshot 2026-09-17 at 10 26 48 AM

SonAndrea:
> I don't think we typically have red buttons, compare the Save/Cancel and Comment/Discard pairs using white and blue. I think confirm should be blue and cancel should be white following this style - let me know what you think. Screenshot 2026-09-17 at 10 23 29 AM Screenshot 2026-09-17 at 10 26 48 AM

I've gone ahead and made the buttons according to other design decisions made on the site for the sake of consistency. I would like to mention that DAILP doesn't really have many strong deletion effects and there is nuance between deletion and discarding. Maybe, as we add more features that permanently alter data, we may want to change our design conventions to delineate a permanent action on data from the database vs an action done in session.

jae-mess:
> > I don't think we typically have red buttons, compare the Save/Cancel and Comment/Discard pairs using white and blue. I think confirm should be blue and cancel should be white following this style - let me know what you think. Screenshot 2026-09-17 at 10 23 29 AM Screenshot 2026-09-17 at 10 26 48 AM
> 
> I've gone ahead and made the buttons according to other design decisions made on the site for the sake of consistency. I would like to mention that DAILP doesn't really have many strong deletion effects and there is nuance between deletion and discarding. Maybe, as we add more features that permanently alter data, we may want to change our design conventions to delineate a permanent action on data from the database vs an action done in session.

Great insight, Andrea. I think that's a really good idea.

PR #786, opened 2026-06-08
#746 Manage Site Pages interface

### LINKED ISSUE #746: Ability to see all pages
Add a "list all pages" option to the admin tools

nole2701:
The link to the page is in the Admin Tools tab on the dashboard (accessible by Admins and Editors).

Features include:
- Lists all site pages
- Searching using page titles
- Filtering by page location
- Enter editing mode for pages
- Delete pages (with confirmation step)
- Create new pages

### DIFF
--- a/graphql/src/query.rs
+++ b/graphql/src/query.rs
@@ -960,6 +960,15 @@ impl Mutation {
             .await?)
     }
 
+    #[graphql(guard = "GroupGuard::new(UserGroup::Editors)")]
+    async fn delete_page(&self, context: &Context<'_>, id: Uuid) -> FieldResult<Uuid> {
+        Ok(context
+            .data::<DataLoader<Database>>()?
+            .loader()
+            .delete_page(&id)
+            .await?)
+    }
+
     // dennis todo: should be admin, but admin accs not implemented yet
     #[graphql(guard = "GroupGuard::new(UserGroup::Editors)")]
     async fn update_menu(&self, context: &Context<'_>, menu: MenuUpdate) -> FieldResult<Menu> {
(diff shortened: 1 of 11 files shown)


types/queries/delete_page.sql
```diff
@@ -0,0 +1,2 @@
+-- delete a page given its id
+delete from page where page_id = $1 returning page_id;
```
sarahkelley42, line 2: To confirm: how permanent do we want deletion to be? Do we assume our users understand the permanence, and/or do we want any concept of undo?

graphql/src/query.rs
```diff
@@ -960,6 +960,15 @@ impl Mutation {
             .await?)
     }
 
+    #[graphql(guard = "GroupGuard::new(UserGroup::Editors)")]
+    async fn delete_page(&self, context: &Context<'_>, id: Uuid) -> FieldResult<Uuid> {
```
sarahkelley42, line 964: For potentially-destructive operations, do we have anything that logs something like "user x took action y"? I could see that being important for auditability and disaster recovery.

sarahkelley42 (commented):
LGTM - does the job. I think there's some opportunity for thinking about disaster recovery (either from user error, attackers, or bugs) when it comes to these operations. But that may be scope creep for what we're trying to accomplish right now.

types/queries/delete_page.sql
nole2701: Right now there's no concept of undo. Deletion is final, but there's a "Confirm Delete Page" check as a 2nd layer to make sure users don't accidentally delete pages by misclicking the delete button.

graphql/src/query.rs
nole2701: Not right now, but I think this as well as undos could be a new ticket in the future. Probably around the time we start working on the Recent Activity dashboard tab which is where actions will have to start being recorded.

PR #787, opened 2026-06-19
#781-782 Audio uploads format and metadata

### LINKED ISSUE #781: Audio file format encoding
The audio recordings done through the site are being saved without a format. The code needs to be updated to clearly and consistently encode to MP3 format

### LINKED ISSUE #782: Add audio metadata
Need to add more descriptive metadata to audio files when audio is being uploaded. This should happen at three levels:
- Actual recording, Audio file, S3 object
- Will need file format and date of upload for all three
- S3 object also needs Relational information (document, speaker, word)

nole2701:
The tasks were to ensure site audio recordings were encoded as mp3 before upload, and add descriptive metadata such as codec/format and relational ids (user, document, word ids, upload date) to the S3 object, and format info to the mp3 file itself.

Changes Summary:
1. Transcoding layer using [lamejs](https://www.npmjs.com/package/@breezystack/lamejs). Decodes the browser's Opus recording to PCM and re-encodes to mp3 (at 128 kbps).
2. Added `metadata` as a parameter for the S3 `PutObjectCommand`, `uploadContributorAudio` writes `format`, `upload-date`, and relational ids. `useAudioUpload` accepts `relations` grabbed from the parent panels (word/document), with the speaker taken from the signed-in user.
3. In `record.tsx` wrap the encoded bytes in a named file with `timestamp` and `.mp3` extension so the file carries its format and upload date at the file level.

### DIFF
--- a/website/src/utils/encode-mp3.ts
+++ b/website/src/utils/encode-mp3.ts
@@ -1,0 +1,59 @@
+import { Mp3Encoder } from "@breezystack/lamejs"
+
+// Samples per frame for MP3 encoding. 1152 is the standard for MPEG-1 Layer III
+const SAMPLES_PER_FRAME = 1152
+// Should be fine for most use cases, but can be adjusted if needed
+const BITRATE_KBPS = 128
+
+export async function encodeBlobToMp3(blob: Blob): Promise<Blob> {
+  // Get raw bytes from the Blob
+  const buf = await blob.arrayBuffer()
+
+  // Decode step and release audio context resources after decoding
+  const audioCtx = new AudioContext()
+  const audio = await audioCtx.decodeAudioData(buf)
+  await audioCtx.close()
+
+  // Metadata for MP3 encoding
+  const channels = audio.numberOfChannels
+  const sampleRate = audio.sampleRate
+
+  const encoder = new Mp3Encoder(channels, sampleRate, BITRATE_KBPS)
+
+  // Convert float audio data to 16-bit PCM as per LAME requirements
+  const left = floatTo16BitPCM(audio.getChannelData(0))
+  // If stereo, get the right channel; otherwise, set it to null
+  const right = channels > 1 ? floatTo16BitPCM(audio.getChannelData(1)) : null
+
+  // Encode audio data in chunks to MP3 format based on the defined samples per frame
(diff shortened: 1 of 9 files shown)


website/src/utils/encode-mp3.ts
```diff
@@ -0,0 +1,59 @@
...
+    const leftChunk = left.subarray(i, i + SAMPLES_PER_FRAME)
+    // If stereo, encode both channels
+    // If mono, the left channel is mirrored to the right channel for encoding
+    const encoded = right
+      ? encoder.encodeBuffer(
+          leftChunk,
+          right.subarray(i, i + SAMPLES_PER_FRAME)
+        )
+      : encoder.encodeBuffer(leftChunk)
+    if (encoded.length > 0) mp3Chunks.push(encoded)
```
sarahkelley42: What should happen if the returned value is zero? Is this a possibility that the lamejs docs say should be checked?

sarahkelley42: formatting nit: I'd generally recommend including the braces even if they're technically optional.

website/src/utils/encode-mp3.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const flushed = encoder.flush()
+  if (flushed.length > 0) mp3Chunks.push(flushed)
```
sarahkelley42: Similar question as above: what's the "happy path" here? If the condition is false, is this indicative of an error, or is this a normal case?

sarahkelley42 (commented):
Makes sense to me. A couple nitpicky questions about the code.

I'm curious how we evaluated the new dependency? Is it a reputable package with reliable maintenance?

I'm also wondering about existing audio; do we need to change anything to handle audio that's been uploaded before this conversion was added?

website/src/utils/encode-mp3.ts
nole2701: If `encoded.length === 0` then it means there were not enough samples to form a full frame so it would just be skipped this call and retained for the next call.

website/src/utils/encode-mp3.ts
nole2701: If `flushed.length === 0` then it means there was no partial frame to be added anyways so skipping it is not an issue.

PR #791, opened 2026-06-24
653 pull format input from iiif

### LINKED ISSUE #653: Pull “Format” input from IIIF
Change Format field to be un-editable for contributors and auto-pull in the format from IIIF.

SonAndrea:
Created a query to generate the IIIF link of the associated opened document to then extract the source type from it's json information. This is auto filled in the format field when editing a document's metadata.


alisony755 (commented):
This worked well when I tested it! I was wondering if the IIIF format should also automatically appear in Document Info? On my end, the IIIF format appears in the editing modal automatically, but the Document Info doesn't show it until the changes are submitted.

SonAndrea:
> This worked well when I tested it! I was wondering if the IIIF format should also automatically appear in Document Info? On my end, the IIIF format appears in the editing modal automatically, but the Document Info doesn't show it until the changes are submitted.

Thanks for the catch; I've gone ahead and made changes to fix this.

website/src/pages/documents/edit-document-modal.tsx
```diff
@@ -239,6 +239,29 @@ export const EditDocumentModal: React.FC<EditDocumentModalProps> = ({
...
+    fetch(iiifData.iiifSourceForDocumentMetadata)
+      .then((res) => res.json())
+      .then((info) => {
+        const formats = info?.profile?.[1]?.formats
+        if (Array.isArray(formats) && formats.length > 0) {
+          setIiifFormat(formats[0])
+        }
+      })
+      .catch((err) => console.error(err))
+  }, [iiifData])
```
alisony755, line 285: Since this code is reused in document-info.tsx, you could extract it into a custom hook (potentially putting it in a separate file in website/src/hooks) and import it in both files

website/src/pages/documents/document-info.tsx
```diff
@@ -42,6 +44,29 @@ export const DocumentInfo = ({ doc }: { doc: Document }) => {
...
+    pause: !doc.id,
+  })
+
+  const [iiifFormat, setIiifFormat] = useState<string>("")
+
+  // When URL is accessible, grab format type from IIIF url info.json
+  useEffect(() => {
+    if (!iiifData?.iiifSourceForDocumentMetadata) return
+
+    fetch(iiifData.iiifSourceForDocumentMetadata)
```
alisony755, line 61: Since `fetch()` doesn't reject HTTP errors like 404/500, `catch()` wouldn't handle unsuccessful HTTP responses. So, you should add a check for `if(!res.ok)` before calling `res.json()`

website/src/pages/documents/document-info.tsx
```diff
@@ -232,7 +257,7 @@ export const DocumentInfo = ({ doc }: { doc: Document }) => {
         <div className={styles.field}>
           <div className={styles.label}>FORMAT</div>
           <div className={styles.value}>
-            {docData.format?.name || "Format Not Yet Available."}
+            {docData.format?.name || iiifFormat}
           </div>
```
alisony755, line 267: I would still have {docData.format?.name || iiifFormat || "Format Not Yet Available."} in case the iiiFormat isn't available or fails to be fetched

SonAndrea: I have the state defaulted to "Format Not Yet Available.", so this will be displayed as a backup.

website/src/pages/documents/edit-document-modal.tsx
SonAndrea: Since it's only duplicated twice and based on previous feedback regarding the limitation of code extraction unless necessary, I think a hook would be a bit overkill to make a separate hook for this pr; however, I do agree that this would be good practice.

website/src/pages/documents/document-info.tsx
SonAndrea: I've gone ahead and added this check, thanks!

website/src/pages/documents/document-info.tsx
alisony755: I'm seeing this for some documents:
image

I think it's because you have:
```
const [citation, setCitation] = React.useState<string>(
    "Format Not Yet Available."
)
```

instead of something like:
```
const [format, setFormat] = React.useState<string>(
    "Format Not Yet Available."
)
```

alisony755 (changes requested):
Thanks for adding the `!(res.ok)` check! I would just double check that the "Format Not Yet Available." check is set correctly/

SonAndrea:
> Thanks for adding the `!(res.ok)` check! I would just double check that the "Format Not Yet Available." check is set correctly/

Thanks for catching this. I must have accidentally deleted it.

alisony755 (approved):
This looks good!

PR #793, opened 2026-06-26
Complete UX for changing between systems

### LINKED ISSUE #761: UX for Parsing System Delineations
Separate the 3 word part systems out and allow users to toggle between which they want to use. 
Consider: Where they will switch between the systems (word pane, user preferences, etc)

Wireframe: https://design.penpot.app/#/workspace?team-id=458ded29-7ef8-80f7-8006-fbe6a1e610a7&file-id=50c86e39-eb77-80af-8007-d1c9a4b75ac7&page-id=50c86e39-eb77-80af-8007-d1c9a4b75ac8&board-id=5b88a21b-8698-808c-8007-d7e9c2b7cf67'

SonAndrea:
**Please ensure word parts are in your database initially for some words such that when you migrate types, they are defaulted to TAOC.**

1. Word parts can now be add in CRG and Learner
2. In the database, word parts and abstract word parts can now be differentiated between systems.
3. In the word pane, you are able to change your system
4. When a word part is added, it is added only for that system
5. Switching between systems only shows the word parts added while in one system


SonAndrea:
reopened by mistake. Will return to this pr in order to fix database changes discussed with Naomi.

doc/database/words.md
```diff
@@ -33,21 +33,34 @@ A join table linking user audio contributions to words in documents. This is a m
 | `include_in_edited_collection` | `boolean`             | Should audio be revealed to readers? Defaults to `false`.               |
 | `edited_by`                    | `uuid? -> dailp_user` | Last Editor to decide if audio should be included in edited collection. |
 
+## `representation_system_type`
```
GracefulLemming, line 36: Would it be possible to use the existing `abbreviation_system` table for this to allow future admin editing?

doc/database/words.md
```diff
@@ -79,20 +92,22 @@ Each system has a different mapping of internal morpheme glosses to display morp
 
 ## `abstract_morpheme_tag`
```
GracefulLemming, line 93: To prevent the team having to retrace what these tables do like we did earlier this year, can you provide a brief description for each of these tables?

types/migrations/20260608232114_add-representation-system-to-morpheme-tags.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE representation_system_type AS ENUM ('CRG', 'LEARNER', 'TAOC');
```
GracefulLemming, line 1: As above, can we use `abbreviation_system` instead of this enum? We can also migrate toward that goal later if you think that grants us more security as we migrate our word parts overall. Let me know what you think.

GracefulLemming (changes requested):
Hi Andrea, here's a partial review for this week mostly focused on documentation and UX. I will finish reviewing files next week. 

For now, you have two major interface issues to address:
1. Switching systems within the word pane closes the word pane, interrupting user workflows.
2.  Adding a new morpheme tag (in my case, under CRG) creates entries for both GLOSSNAME and CUS:GLOSSNAME. Both of these are shown to the user via the word part dropdown, which will be confusing if unaddressed.

Please also provide thoughts on the use of your enum vs the abstract morpheme tag table for `representation_system_type`

doc/database/words.md
SonAndrea: Will do when I make changes for this pr.

types/migrations/20260608232114_add-representation-system-to-morpheme-tags.sql
SonAndrea: The abbreviation system logically makes the most sense to use and initially tried to use it for the additional features other than the current saving morpheme pipeline that relies on the abbreviation_system table; however, when working on the pr, I struggled to make it work due to some constraints put on either table. I didn't document it and opted to use the enum as a quicker solution that limited the number of steps taken. I'll revisit it to see if I an rewire the tables to use the abbreviation_system ids instead of the enum.

doc/database/words.md
SonAndrea: It certainly is possible to be done but based on how this pr was done, I am considering removing this enum. Where these values were used, they will be replaced with the abbreviation_system being used. Using the abbreviation_system table, it can be utilized for future admin editing.

SonAndrea:
> Hi Andrea, here's a partial review for this week mostly focused on documentation and UX. I will finish reviewing files next week.
> 
> For now, you have two major interface issues to address:
> 
> 1. Switching systems within the word pane closes the word pane, interrupting user workflows.
> 2. Adding a new morpheme tag (in my case, under CRG) creates entries for both GLOSSNAME and CUS:GLOSSNAME. Both of these are shown to the user via the word part dropdown, which will be confusing if unaddressed.
> 
> Please also provide thoughts on the use of your enum vs the abstract morpheme tag table for `representation_system_type`

1. This is a re-rendering issue caused by the local storage being updated in the system changing. I'll see if I can make changing systems prevent the request.
2. I'll remove the creation of custom tags and only use them as fallback which was the intial intent.

Regarding the thoughts of using my enum vs the table, I outlined it in one of my comments.

"The abbreviation system logically makes the most sense to use and initially tried to use it for the additional features other than the current saving morpheme pipeline that relies on the abbreviation_system table; however, when working on the pr, I struggled to make it work due to some constraints put on either table. I didn't document it and opted to use the enum as a quicker solution that limited the number of steps taken. I'll revisit it to see if I an rewire the tables to use the abbreviation_system ids instead of the enum."

But as I noted, I'll go back to keep it consistent as this  was done to get around having to include another table reference as I struggled to get it to work.

PR #804, opened 2026-07-09
Hidingpublishing frontend

### LINKED ISSUE #783: Hiding/Publishing for Collections
Need a feature so that collections can be worked on without appearing on the site to others. This should be similar to page publishing where content can be worked on by those with correct access levels, and then published for all to see.

**Screenshot of Jae's mockup below**

SonAndrea:
Collections are defaulted as hidden. 
1. Editors and Admins can see all hidden and published collections from the main page and can toggle visibility
2. Contributors can see all hidden and publish collections but cannot toggle visibility
3. All other users can only see published collections


website/src/components/collection-card.css.ts
```diff
@@ -86,3 +84,63 @@ export const actionButton = style({
     transform: "scale(1.02)",
   },
 })
+
+export const toggleButton = style([
+  button,
+  {
+    position: "absolute",
```
GracefulLemming, line 96: Did you mean `position: "static"` here?

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position

website/src/components/collection-card.tsx
```diff
@@ -1,33 +1,61 @@
...
 
 export const CollectionCard = (props: {
   thumbnail: string
   header: { text: string; link: string | undefined }
   description: string
   buttonLabel: string
+  collectionId?: string
+  isHidden?: boolean
+  canToggle?: boolean
+  canView?: boolean
```
GracefulLemming: Could you share why you introduced the `canView` prop?

If the parent list for cards is responsible for rendering cards (as in `index.page.tsx:62`), why would a collection card need to be aware of its own visibility? 

Additionally, isn't the possibility of viewing a particular card the UX realization of the hidden status of a collection and a user's role? If so, I think this field would be redundant since we already can access a user's role and `collection.isHidden` within this component.

In general, I think it's a good idea to keep user-branching logic as encapsulated as possible. To me, this prop lifts some user-branching logic out of this component which  makes the rendering logic for individual cards more opaque. 

Consider the following interpretations of line 43:
- when this card can be toggled and can be viewed, show the toggle badge.
- when this card can be toggled and the user is not a reader, show the toggle badge.
Which of these is more clear? Does the code reflect that?

GracefulLemming: I'm also wondering if the same concerns apply to `canToggle`.  If lifting these fields is due to DRY concerns, then we should make our user context provider more robust rather than passing user role conditions through props.

Please make the following assertions more clear:
1. If a user is a reader,  do not show the hidden icon nor the toggle badge
2. If the user is a contributor, show the hidden icon but not the toggle badge
3. If the user is an editor or admin, show both the hidden icon and the toggle badge

GracefulLemming (changes requested):

website/src/components/collection-card.css.ts
SonAndrea: I used absolute because I want the button to appear as floating in the top right above all elements and to not affect the rest of the card. When I used static initially, the button doesn't behave as intended.

website/src/components/collection-card.tsx
SonAndrea: Regarding line 43:
"When this card can be toggled and the user is not a reader, show the toggle badge."
This is a more clear interpretation (and is what this pr implements) but in quite a convoluted way which I'm now realizing after your pushback. 

Regarding redundant user check props:
Right now, canView refers to if it is a contributor who can only view the card and not edit it. This is used to display the hidden or published badge but delineates from editors who can see this but also the toggle button; however, I've been making this much more complicated than needed. Both canToggle and canView don't need to be passed as props and I'll make changes to reflect this.

Regarding redefining my assertions:
1. If a user is a reader, the user should not be able to view a hidden collection card on the home page or access a hidden collection. For published collections, they can view them as currently seen on DAILP.
2. If the user is a contributor, the user can view both hidden collection cards and published collection cards. They can access the collections regardless of if it is published so they can make contributions at any time.
3. If the user is an editor or admin, they can view and access both hidden and published collections just as contributors can. In addition, they are able to toggle any given collection from the home page, toggling the collection from a hidden state and a published state.

website/src/components/collection-card.css.ts
GracefulLemming: Screenshot 2026-07-22 at 12 24 23 PM

FYI I am seeing this on my dev server (Safari 18.5), I figured the static label may be the cause but Im not certain. 

Its also worth knowing that the cards still have the icons in the right place:
Screenshot 2026-07-22 at 12 33 53 PM

GracefulLemming: update: I think the extra overlapping show/hide buttons are related to the hover behavior for the hide and publish buttons. Essentially, on safari the yet to be used buttons jump to the top right of the page for some reason.

GracefulLemming: This issue also impacts the deploy preview– this is now a blocking issue.

GracefulLemming (changes requested):
Thank you for these changes @SonAndrea! I love that you extended our guard pattern to include a dedicated guard for collections. 

There is one minor formatting bug, but it is blocking nonetheless. [See my notes in our existing discussion](https://github.com/NEU-DSG/dailp-encoding/pull/804#discussion_r3633236346). You're in the home stretch! Once this bug is addressed we can merge.

@LN7-cyber and @HaileyPunis:
I confirmed in the deploy preview that documents and editorial content within an unpublished collection cannot be seen by readers (only editors and contributors). 

I also tested what happens when a document is in both a published and unpublished collection:
1. Setup: Open deploy preview, log in as editor. Scroll down on the homepage to publish CWKW and hide WJS.
2. Test: In an incognito tab (reader, not logged in), access the [Story of Millie Pigeon in WJS](https://deploy-preview-804--dailp.netlify.app/collections/willie-jumper-stories/millie-pigeon) and [in CWKW](https://deploy-preview-804--dailp.netlify.app/collections/cwkw/story-millie-pigeon/)
3. Results: CWKW loads the story, WJS shows the message "Please log in to access this page."
4. Interpretation: A collection-level guard on a document can be circumvented by accessing that document from another collection.
5. Possible interventions: (a) roll out static/point-in-time representations of documents to better moderate what is shown when, and (b) roll out document-level show/hide. (a) has been a long-term goal of ours for some time; (b) is a lower effort stepping stone toward (a).

SonAndrea:
I'm currently working on seeing if I can resolve this ux bug regarding the button overlap. Additionally, I plan on including a document level show/hide based on if it is in a hidden collection to address the secondary bug. The static/point in time representation does seem a big of a larger undertaking and may be better fit for another pr which I would not be opposed to work on.

SonAndrea:
After taking a look into the provided case of Millie Pigeon, I wanted to know what exactly the expected case should be. There are two database entries for the referenced docs: 

1. /collections/willie_jumper_stories/millie_pigeon
2. /collections/cwkw/story_millie_pigeon

Trying to access either of the following two results in a failed page being accessed:

1. /collections/willie_jumper_stories/story_millie_pigeon
2. /collections/cwkw/millie_pigeon

I was just curious as exactly how the collection guard is being circumvented given that it appears to me that these are duplicate documents where one is visible in CWKW but not visible in WJS in your given environment case. Do you mean that both stories should be hidden given that they are duplicated and one is hidden so the other should be too?

GracefulLemming (approved):
Thanks for the fix, feel free to merge!

PR #807, opened 2026-07-14
Call turnstile siteverify sarahs version

### LINKED ISSUE #731: Turnstile Implementation
Original PR was #525

sarahkelley42:
Adds a Lambda which calls Turnstile. The Lambda handling the main HTTP endpoint does not have outbound network access, so the new Lambda exists to give the graphql endpoint a controlled way to call it instead. Adds Terranix definition of the lambda and associated policy.


GracefulLemming:
Hi Sarah, thank you for this work so far. As discussed over email, I agree that dependency management is a larger issue we need to tackle together. I appreciate you separating that concern out of this work to the furthest extent possible.

In my eyes, here are our next steps:
1. Implement logic marked with the `todo!` macro in `query.rs:validate_turnstile_token(...)`: (a) dispatch validation to outbound lambda using lambda's invoke feature, and (b) processing and returning the validation result for both local and lambda deployment environments.
2. Write terranix code configuring outbound lambda function. I can provide some starter code if you would like.
3. Merge this PR, moving changes from your fork to its corresponding feature branch on the daily-encoding repo
4. Ensure formatting and logic checks pass
5. Full code review and manual test– we should plan to do the manual testing together on Zoom
6. Merge to main

sarahkelley42:
Sounds good. For the terranix code, if you've got work in progress that would help, but if not I'm not too worried and I can take a crack at it.

-------- Original Message --------

sarahkelley42:
I think this is ready for review. I'm noting that the Terraform Plan step is failing; it's unable to get credentials, which I presume are stored as secrets only accessible to the main dailp-encoding repository; since this is a PR from a fork I'm guessing that's why it doesn't have access. (Rightly so; we don't want that exposed to code running from forks.)

@GracefulLemming Looks like I don't currently have the necessary access to push this branch to the dailp-encoding repo; I'd like to confirm that the CI checks pass before we merge to main. To that end I've pointed this PR at the original turnstile branch instead of main; if we merge to the branch we should see the PR checks passing on the original: https://github.com/NEU-DSG/dailp-encoding/pull/792

graphql/src/query.rs
```diff
@@ -985,22 +988,50 @@ impl Mutation {
...
+            dailp_graphql::service_integrations::turnstile::validate_token(token)
+                .await
+                .map_err(|_| dailp::async_graphql::Error::new("Failed to validate Turnstile token"))
         } else {
             use aws_sdk_lambda as lambda;
             let config = aws_config::load_from_env().await;
             let client = lambda::Client::new(&config);
+            let function_name = match std::env::var("OUTBOUND_LAMBDA_NAME") {
+                Ok(name) => name,
+                Err(_) => return Err("Failed to validate Turnstile token".into()),
```
GracefulLemming: Is this the most precise error we can give here?

It seems like it would be helpful to flag this as a configuration error since this is catching when `std::env::var(...)` fails, not the downstream call to turnstile's API.

GracefulLemming (changes requested):
Thank you for this work @sarahkelley42, I have one minor note about logging for you to address. The terraform code looks good from your fork, I suggest we merge this branch into the DAILP version of this branch before our meeting so we can talk through what will happen on merge to main.

graphql/src/query.rs
sarahkelley42: Agreed, I added a log line in this branch. I think it makes sense to leave the error we return to users vague, so that we expose less information about our infra to would-be attackers, and add logging so that we can see what's going on afterward.

sarahkelley42:
@GracefulLemming  I made the requested change. I don't have the requisite permissions to merge or push into the DAILP repo, so if you want this merged in I think you'll have to be the one to push the button. :)

PR #818, opened 2026-07-22
Added upload audio button for documents

### LINKED ISSUE #816: Upload Document Audio From Device
Additional option to upload audio from device for overall document audio.

SonAndrea:
When a user is logged in, and can view the audio recording button for an overall document audio, they are given the option to upload audio instead. This will give user's the option to upload audio from their device which can then be uploaded as normal.


GracefulLemming (approved):
Approving, but...

I found one UX issue:
1. users must allow microphone access in order to see the upload button

I also see you are modifying the recording component directly. Is there a reason you avoided the existing upload code held in the same folder?

Im flagging these issues as non-blocking here so we can iterate on them while still being able to share with the team tomorrow.

PR #822, opened 2026-07-23
809 metadata addition dates field

### LINKED ISSUE #809: Metadata Addition "Dates" field
- Implementation will require a table to link “Dates” metadata and to the documents. 
- This is meant to be dates WITHIN the story. We should use it as "Temporal Coverage" which is part of Dublin Core instead of just labeling it "Date".

alisony755:
- Added key dates field to metadata display and editing


types/queries/insert_document_key_dates.sql
```diff
@@ -0,0 +1,3 @@
+-- Insert the new set of key date IDs
+insert into document_key_date (document_id, key_date_id)
+select $1, unnest($2::uuid[]);
```
GracefulLemming, line 3: Can you explain what you are trying to accomplish with this line?

website/src/pages/documents/edit-document-modal.tsx
```diff
@@ -95,6 +96,8 @@ const approvedSpatialCoverages = [
   "Dubai, UAE",
 ]
 
+const approvedKeyDates = ["February 12, 1964"]
```
GracefulLemming: Is there a reason this date is hard-coded?

GracefulLemming (changes requested):

GracefulLemming (commented):
Hi Alison, I think this looks good overall but I have one SQL and one web question for you to answer before approval.

types/queries/insert_document_key_dates.sql
alisony755: This line inserts a row into document_key_date for each key date ID, associating all of those key dates with the document when updating metadata

website/src/pages/documents/edit-document-modal.tsx
alisony755: I was getting an error if I left it as `const approvedKeyDates = ["February 12, 1964"]`, but I could change it to `const approvedKeyDates: string[] = []`

PR #823, opened 2026-07-23
Remove Terranix definition of Lambda:Invoke policy

sarahkelley42:
We are creating a new Lambda to allow us to call the Turnstile API. The main GraphQL Lambda needs to be able to invoke this function, so we will need an IAM policy to allow this action. However, the role used by our infra deployments is not allowed to all the iam:PutRolePolicy. This change deletes the policy from our Terranix code while keeping the lambda intact; we will be able to invoke the lambda by hand elsewhere, but the GQL lambda will not.

Once we have the necessary trust policy in place, this commit can be reverted.

### DIFF
--- a/graphql/src/query.rs
+++ b/graphql/src/query.rs
@@ -995,43 +995,14 @@ impl Mutation {
                 .await
                 .map_err(|_| dailp::async_graphql::Error::new("Failed to validate Turnstile token"))
         } else {
-            use aws_sdk_lambda as lambda;
-            let config = aws_config::load_from_env().await;
-            let client = lambda::Client::new(&config);
-            let function_name = match std::env::var("OUTBOUND_LAMBDA_NAME") {
-                Ok(name) => name,
-                Err(_) => {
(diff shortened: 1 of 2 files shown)


sarahkelley42:
@GracefulLemming The policy we're looking for would look like this:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "invoke-outbound-turnstile",
      "Effect": "Allow",
      "Action": "lambda:InvokeFunction",
      "Resource": "arn:aws:lambda:*:ACCOUNT_ID:function:dev-outbound-turnstile*"
    }
  ]
}
```

Note that ACCOUNT_ID would need to be filled in, and dev-outbound-turnstile would need to be environment-specific. If we want this to be the same across all environments (I might recommend against this), we could change the resource to all wildcards e.g. `"arn:aws:lambda:*:*:function:*my_lambda_function*"`. More wildcards = fewer updates but a wider set of allowed actions.

GracefulLemming (approved):

PR #832, opened 2026-07-24
minor fixes to fix at dev-check

SonAndrea:
Fixed merge conflict and undefined check that was not flagged during compilation but preventing further pushes.


GracefulLemming (approved):
I notice these changes are duplicated in #833. Either merge this first or close it as stale.

PR #833, opened 2026-07-24
Upload audio changes

### LINKED ISSUE #830: Adjustments to Uploading Audio
Explained reasoning to directly modifying record.tsx and fixing minor bugs.

SonAndrea:
Modified the audio component to allow a user to upload audio when the microphone is disabled. Additionally, in the original pr that was merged, @GracefulLemming, asked:

"I also see you are modifying the recording component directly. Is there a reason you avoided the existing upload code held in the same folder?"

I used and modified the existing upload code and put it directly within the audio recording as I find that typically on other sites with similar features, the record and upload are within the same component. With using the current implementation in upload.tsx, there would be two separate sections where users can do both at the same time. I wanted to limit the freedom to simplify it and prevent any complications of a user being able to submit uploaded and recorded audio in different places.


GracefulLemming (approved):
Thank you for these changes @SonAndrea and thank you for explaining why you didn't use the existing upload component. I see that users now do not need to allow their microphone access to upload audio, and I see you updated some of the prompt text for browsers that don't support audio recording. 

Based on your comments, it might be helpful if you remove the old upload code so we can cut back on clutter but that can be done in a separate PR. For this PR, we should merge to main.

PR #844, opened 2026-08-07
Database bulk export– tabular

GracefulLemming:
1. Exports data from the database and saves them as a CSV or pg_dump file.
2. Uploads files to S3
3. Recovers from recoverable errors where applied
4. Generates logfiles indicating where failures occured


scripts/src/file_utils.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
+
+#######################################
+# Create a file with a given header/content, creating parent directories
```
SonAndrea, line 4: This function looks pretty exhaustive and good to me. Only thing I could think of is providing some output in the event that this function creates new directories upon file creation (this is more targeted for manual use cases of accidental directories being made).

PR #847, opened 2026-08-09
Create VPC terraform module

sarahkelley42:
This is a work in progress. The main application depends on having a VPC in place, and takes the VPC and subnet IDs as inputs. Today these are provided by ITS but future sites will need to be able to create these resources independently.

This work is incomplete; we also need to create a bucket for this module's state, and we will need to ensure that we can pass the outputs from this module to the inputs of the main application module for smooth deployments.

Finally, note that this is currently written in HCL instead of Terranix. These should be isomorphic, so we can convert the HCL to terranix later, but I'm more immediately comfortable working with HCL so I'm starting there and can convert when this is in a more stable place.

Tested by creating the resources in a personal AWS account.

### DIFF
--- a/.terraform.lock.hcl
+++ b/.terraform.lock.hcl
@@ -5,6 +5,7 @@ provider "registry.terraform.io/hashicorp/aws" {
   version     = "6.2.0"
   constraints = ">= 2.0.0, >= 2.55.0, >= 3.0.0, <= 6.2.0"
   hashes = [
+    "h1:Li35IiZOPaw01q7wZmvj1NiUnGUHqX6phrx9OWXU4eI=",
     "h1:ziUjk3KGwBa7bAwZaPuQklfcQ3Qlx2d0rlYFvdZn4g8=",
     "zh:224b20371f7c7ce14d69a84e16ae94baa0a06c132474d4bc4d192d86936bc750",
     "zh:2c079ad275c32b9abae7616d07c24901340207a85995ce0025ac38af16b317b7",
--- a/flake.nix
+++ b/flake.nix
@@ -65,8 +65,8 @@
         in naersk.buildPackage {
           root = ./.;
           src = packageSrc;
-          doCheck = true;
-          doTest = true;
+          # doCheck = true;
+          # doTest = true;
 
           nativeBuildInputs = [ cc ];
 
--- a/terraform/auth.nix
+++ b/terraform/auth.nix
@@ -5,7 +5,7 @@ in {
   config.resource = {
     # User Pool Setup
     aws_cognito_user_pool.main = {
-      name = "dailp-user-pool";
+      name = "bogus-user-pool";
       username_attributes = [ "email" ];
       auto_verified_attributes = [ "email" ];
       admin_create_user_config.allow_admin_create_user_only = false;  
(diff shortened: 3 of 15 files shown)


sarahkelley42:
@GracefulLemming The main thing I'm taking note of is that we have lots of uses of `getEnv` in the main application; my recommendation is that we transition these to using Terraform's variable system. Two reasons: 1, this allows us to separate the build artifact (a terraform JSON) from the deployment configuration (which includes runtime config + secrets); and 2, this opens some doors for us to e.g. reference the outputs from this module as inputs in the main app module.

note that I have not yet gotten to the point of fully deploying the app to my account; I'm still tracking down a couple missing dependencies that we need as a base.

sarahkelley42:
FWIW I think the only remaining piece I need to set up in this is the bastion host. I'll want to check the most sane way to handle this. We only want resources to be owned by one terraform module, so I don't think it makes sense to move the bastion definition to the base infra module. However we'll want to be able to skip importing if the resource doesn't yet exist.

For now, I'm simply commenting out the imports so I'll end up creating them from scratch on my account.

```
$ terraform plan
module.bastion_host.data.aws_iam_policy_document.main: Reading...
module.bastion_host.data.aws_iam_policy_document.main: Read complete after 0s [id=REDACT]
module.bastion_host.data.aws_region.default: Reading...
module.bastion_host.data.aws_region.default: Read complete after 0s [id=us-east-1]
module.bastion_host.data.aws_iam_policy_document.default: Reading...
module.bastion_host.data.aws_iam_policy_document.default: Read complete after 0s [id=REDACT]
module.bastion_host.aws_iam_role.default[0]: Preparing import... [id=dailp-dev-bastion]
module.bastion_host.aws_iam_role.default[0]: Refreshing state... [id=dailp-dev-bastion]
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.media_storage,
│   on config.tf.json line 854, in resource.aws_s3_bucket.media_storage:
│  854:         "acl": "private",
│ 
│ Use the aws_s3_bucket_acl resource instead
│ 
│ (and 8 more similar warnings elsewhere)
╵
╷
│ Error: Cannot import to non-existent resource address
│ 
│ Importing to resource address module.bastion_host.aws_iam_role_policy.main[0] is not possible, because that address does not exist in configuration. Please ensure that the resource
│ key is correct, or remove this import block.
╵
╷
│ Error: Cannot import to non-existent resource address
│ 
│ Importing to resource address module.bastion_host.aws_instance.default[0] is not possible, because that address does not exist in configuration. Please ensure that the resource key is
│ correct, or remove this import block.
```

sarahkelley42:
Using this as an opportunity to catalog things that need to be tweaked to spin up a from-scratch instance. I've pushed some commits with hacks I've put in to get resources to deploy onto my AWS account. I still have a couple to go. The latest apply attempt I have is:

- Need to create a key pair in the bedrock module (or elsewhere)
- Need to clarify what is wrong with the creds config; haven't delved into anything Amplify yet.

```
│ Error: creating Amplify App (dailp-dev): BadRequestException: There was an issue setting up your repository. Please try again later.({
│   "message": "Bad credentials",
│   "documentation_url": "https://docs.github.com/rest",
│   "status": "401"
│ })
│ 
│   with aws_amplify_app.dailp,
│   on config.tf.json line 319, in resource.aws_amplify_app.dailp:
│  319:       }
│ 
╵
╷
│ Error: creating EC2 Instance: InvalidKeyPair.NotFound: The key pair 'dailp-dev-2024' does not exist
│       status code: 400, request id: 9f225f05-5d65-4b89-8dc1-c067102b9d60
│ 
│   with module.bastion_host.aws_instance.default[0],
│   on .terraform/modules/bastion_host/main.tf line 54, in resource "aws_instance" "default":
│   54: resource "aws_instance" "default" {
│ 
```

PR #855, opened 2026-08-12
Enhanced Word Search Feature

### LINKED ISSUE #810: Update Index Search
- Update the index search to hide DocID and surface Collection + Story name.
- Also for the search:
    - Hide unpublished materials from showing in search
     - Top level search needs to be an advanced search where you can toggle by collection + dictionaries.

SonAndrea:
Added the following feature to the word search page:
1. Each entry displays it's title instead of its document ID
2. Each entry now displays the collection it belongs to
3. When logged in as a reader, results from a hidden collection no longer display
4. An enhanced filter menu can be accessed next to the search bar to only display results of selected collections (readers only see non hidden collections in this menu) or dictionaries


migration/src/main.rs
GracefulLemming, line 1: Can you revert this change? It seems unrelated to the current PR.

website/src/components/lexical-search.tsx
```diff
@@ -41,24 +77,95 @@ export const LexicalSearch = () => {
...
+            }}
+          />
+
+          <div className={css.filterWrapper}>
+            <button
+              type="button"
+              className={css.filterButton}
+              onClick={() => setFilterOpen((open) => !open)}
+            >
+              Filter {filterOpen ? "▲" : "▼"}
```
GracefulLemming, line 97: Is there a reason you chose to present these arrows as strings rather than SVG symbols like we do for our other icons?

website/vite.config.ts.timestamp-1786037727892-4d564d2e96a7a.mjs
GracefulLemming, line 1: Never commit this file type. I'll push an update to the gitignore for this.

GracefulLemming (changes requested):
Hi Andrea,
These changes make the search page feel so much more approachable! thank you for the work. I have three minor notes in the comments and one UI request.

It would be helpful if users could see at a glance where an occurrence of a word is within a document from the search page. To that end, could you consider the following additions:
1. For documents with more than one page, show the page number
2. For documents with more than one paragraph, show the paragraph
3. Show the word number

3 is the only hard requirement here and I recognize space is very limited per result, so let me know what you think of the other two. You might consider putting this info between the document and collection title.

migration/src/main.rs
SonAndrea: Thanks for catching this; I was working on this branch when we we're having our discussion Monday.

website/src/components/lexical-search.tsx
SonAndrea: Not particularly; I just didn't check for the fi icons we currently use to match the styling and used a shortcut. I'll standardize it.

website/vite.config.ts.timestamp-1786037727892-4d564d2e96a7a.mjs
SonAndrea: I'll remove it for the next push; unsure why this appeared.

SonAndrea:
I've gone ahead and added the following two cases:
1. For documents with more than one page, show the page number
3. Show the word number

I did not add the paragraph number since it does feel slightly redundant given that page numbers (which I find more helpful) are accessible with the current graphql and that all three feels a bit too claustrophobic.

jae-mess:
Hi Andrea, I'm getting an error when searching both signed in (contributor) and signed out. 
Screenshot 2026-09-17 at 10 35 24 AM

SonAndrea:
> Hi Andrea, I'm getting an error when searching both signed in (contributor) and signed out. Screenshot 2026-09-17 at 10 35 24 AM

I'm unsure why this is the case. Is this screenshot from "[base url]/tools/search/?query=write"? These are my results of the same search:

image

